PR #2, opened 2016-08-18
Adding image data uri support and npm package.json

DzmitryDulko:
Added support for data image uri
Added package.json for consumption from node/npm


DzmitryDulko:
+Updated jsZip library to newer version

gitbrent:
Great work, thanks!

I'm moving the new demo html file to demo_edge so the main demo.html page will work IE and Safari.

PR #16, opened 2016-12-16
Pptxgen es6 poyline

nadhub:
Can create Polyline and Polygone


gitbrent:
The library is slated to support IE11 for a while longer (until the end of 2017 probably), so i'm closing this for now.

ES6 features will be implemented once IE11 support is dropped.

Thanks!

PR #41, opened 2017-02-14
Call back for save method (nodejs only)

ellisgl:


gitbrent:
Thanks @ellisgl !

ellisgl:
@gitbrent no problem.

PR #53, opened 2017-03-09
added image type to shapes to allow images to be placed on top of shapes, added more properties to ppt document

ericwgreene:
When creating masters, images are processed before shapes causing all images to appear below shapes when a shape and images occupy the same location on the slide. Adding a shape type of image in the same manner as text resolves this problem by allowing stacking according to the array order.

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -100,6 +100,10 @@ var PptxGenJS = function(){
 
 	// B: Set Presentation Property Defaults
 	gObjPptx.title = 'PptxGenJS Presentation';
+	gObjPptx.author = 'PptxGenJS';
+	gObjPptx.revision = '1';
+	gObjPptx.subject = '';
+	gObjPptx.company = 'PptxGenJS';
 	gObjPptx.fileName = 'Presentation';
 	gObjPptx.fileExtn = '.pptx';
 	gObjPptx.pptLayout = LAYOUTS['LAYOUT_16x9'];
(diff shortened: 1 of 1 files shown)


gitbrent:
Hi @ericwgreene ,

Thanks for opening this issue/request.  The issue is fixed now, although I'm not pulling this code as i did a more intensive re-write of the master building code that addresses other issues with ordering as well.

Try it now and let me know if it works for you.

The demo page (`pptxgenjs-demo.html`) Masters section shows images over shapes now as well.

ericwgreene:
Can you add support for the additional properties which I included in my second commit?

gitbrent:
Oh, i see.  Because of the way the Masters build their objects in sets (all images, all shapes, etc.) you can't interleave objects and images.

Given that, the larger solution is to modify the way Master Slides ingest objects to a become a single set that can have a mixture of types.

The degree of complexity being utilized in masters by yourself and others makes re-working the master params a worthwhile endeavor, so i'll start work and let you know when i've got something for you to test with.

Thanks!

ericwgreene:
Exactly. Also, I added some additional property settings for company, author, subject, etc...  Can you add those as well?

gitbrent:
Hi @ericwgreene ,

Thanks again for the requests.

Slide Master objects can now be added in any order and build up in a sensible manner.

[Slide Master Examples](https://github.com/gitbrent/PptxGenJS#slide-master-examples) and Demo page updated to show the new methodology.

PR #85, opened 2017-05-15
Adding ROUND_RECTANGLE radius (r) and line dash options

ivolazy:
@gitbrent, I needed to add predictable rectangle rounding radius for rectangle shapes and shape line dashes. Please review and merge.


ivolazy:
Need to add examples

PR #107, opened 2017-06-21
Better client side support

santi-git:
The PR's intent is to improve the experience of using the library on the client side, the changes that were made are the following:

- Uploaded the repository to the bower, so now it can be installed via bower install pptxgen
- Added gulp task to generate a bundle with all the core files needed to run pptxgen (jszip, jquery)

If the PR is merged, the version tag should be bumped and the new version should be uploaded to bower, given there is no contribution instructions I did my best to be us tidy as possible. I don't now how you handle builds internally, or if you do it at all so I tried not to go too far with that.
I'm using pptxgen at work and this improvement would make it easier for me to use it. 

Hope it helps!


gitbrent:
Hi @santi-git ,

Thanks for the PR and doing to bower registration! I've incorporated the changes into the project.

Also, thanks for getting me started with gulp - i've been meaning to setup a workflow for a bundled file for a while now. :-)

PR #128, opened 2017-08-07
Add series data point colors

clubajax:
This commit adds a `<c:dPt>` node, which represents each data point in a series. It's possible there is cruft in the child nodes - this was pulled directly from a modified PPTX.

As per the conditional, if there are colors, and this is not a collection of series (clustered or stacked), it will apply the colors.


clubajax:
What is the corruption? I'm exporting several types of charts. Maybe I missed a scenario.

clubajax:
Oh, I wonder if I did not build the code correctly. I had trouble with the source maps. I was running the unbuilt file, and not using the extras.

PR #136, opened 2017-08-10
Shadow corrections bugfix

kajda90:
PptxGen detects incorrect shadow options values but the correction does not work (invalid PPTX produced). This PT assigns the corrected values to the proper object.


gitbrent:
Thanks!!

kajda90:
You're welcome :)

PR #140, opened 2017-08-11
Features to Enable Tornado Chart

clubajax:
We have a requirement to build a tornado chart. PPT has the ability to do this, but it requires customizing a typical bar chart series. This PR additionally has some general features.

- `catAxisLabelPos`:  By default the labels on the tornado chart will be on the 0 axis line. This allows for `low` (left) and `high` (right)
- `valAxisLineShow`: Setting to false will hide the axis line (defaults true)
- `axisLineColor`: color of axis lines
- `gridLineColor`: color of grid lines (no features were added to handle line styles)
- `invertedColors`: This is a list of colors, much like chartColors. These are applied if the value is less than zero. PPT defaults to white.
- `valueBarColors`: Forces chartColors on multi-data-series. This is kind of a special switch that was specific to tornado charts. Not sure if it has any other use cases.

tornado-chart


gitbrent:
Issue #151 added this code.

Thanks @clubajax !

clubajax:
@gitbrent, your welcome, but I don't see the changes in master?

PR #141, opened 2017-08-11
Charts: layout option

kajda90:
I've noticed that auto-layouting charts by PowerPoint resulted in adding some extra space between chart boundary and plot area. Sometimes, I need the plot area to fulfil the chart absolutely (I expect the chart width to be 1 inch when I set `width` option to 1.0). It can be done by setting layout.
If the `layout` option is not specified, the chart layouting works as before, i.e. it's handled by PowerPoint.


gitbrent:
Thanks @kajda90 !

PR #142, opened 2017-08-12
New feature - doughnut charts

kyrrigle:
Hi Brent.  Thanks for this nifty powerpoint generator.  Very nicely done.

We had a need for creating doughnut charts which turned out to be a simple variation on pie charts.

I've tested locally and have changed one of the pie charts in pptxgenjs-demo.js to be a doughnut.

Matt


gitbrent:
Thanks the contribution @kyrrigle !

Issue #96, opened 2017-05-29
Gridlines and valAxisLabel


Shankar000536:
When can i expect Gridlines and valAxisLabel format options? (Removing Gridlines and valAxisLabel in bar chart?)

thanks,
Shankar

kajda90:
Hi Shankar,
I've create a pull request #143 for this functionality. Now, it's waiting for @gitbrent's approval.

PR #143, opened 2017-08-14
Charts: Optional displaying gridlines and axes

kajda90:
- Hi, I've needed to insert a sparkline chart - no labels and no extra lines. Thus, I've added options that enables authors to hide vertical or horizontal grid lines and similarly even the Y- and X-axis. This was also required in #96.
- Y-axis supports setting its maximum value but not the minimum. Therefore, I've added `valAxisMinVal` option.
- According to the TODO mention in the code, I've implemented styling the gridlines, so now a styling object can be passed to the `valGridLine`/`catGridLine` or `"none"` if should not be visible. (_valGridLine_ is still visible by default unlike _catGridLine_.)


gitbrent:
Hi @kajda90 ,

Thank you very much for the contribution!  These are super helpful chart options.

kajda90:
You're welcome, I'm glad you like it. By the way, issue #96 can be closed now I guess :)

PR #144, opened 2017-08-14
New Feature: Axis Titles 

kyrrigle:
New options to add support for turning on and formatting axis titles.

This is generalized to use same code as chart titles to generate the `<c:title>`
This also allows titleRotate (should anyone want that)
And I spotted a disconnect between the titleFontSize documentation and implementation.  Changed the implementation to match (font size 18)

Readme is updated with options and the "Lots of Bars" demo shows off the options


gitbrent:
Hi @kyrrigle ,

Great work - thanks for the contribution!

Issue #145, opened 2017-08-15
Chart question about rows/columns


kyrrigle:
Hi Brent,

I have a need to plot time series with a relatively small number of series (10 or so) but with a potentially large number of data points (by day for the last 90 days for example).   

Some problems I've encountered- 

1. Cannot have nulls in the data values.  I've got a [patch](https://github.com/lexalytics/PptxGenJS/tree/bugfix/allow-null-values) for this that seems to work.

2. Currently no support for proper dates for the data labels.   PPT/XLS supports this and seems like a straight forward change (`<c:numRef>` instead of `<c:strRef>`)

3.  When I look inside the Excel sheet I see the data is modeled with the series as rows and the data points as columns.   Near as I can tell this will substantially limit how many data points I can have, right?

For example, here's what I get when I open up the Excel sheet of one of the sample line charts:

_ | Q1 | Q2|  Q3|  Q4
-- | -- | -- | -- | --
Red |22 | 11 | 18  |15
Amb |18 | 19|  32 | 19
Grn| 50 | 71 | 5  | 13
Unk| 7|   7|   3|   3

It would seem that instead the series should be columns with the rows holding the data points:

_ |    Red |Amb| Grn| Unk
-- | -- | -- | -- | --
Q1 | 22 | 18 | 50 | 7
Q2  |11  |19 | 71 | 7
Q3  |18 | 32|  5 |  3
Q4  |15|  19  |13|  3

This is also the structure you get by default if on a blank PPT slide you insert a new line chart:

_ |    Series 1 |   Series 2 |   Series 3
-- | -- | -- | -- 
Category 1 | 4.3 |2.4| 2
Category 2 | 2.5| 4.4| 2
Category 3 | 3.5 |1.8| 3
Category 4 | 4.5 |2.8| 5

Before I dig in to patching for this I was curious if there was a reason you went the way you did and if you had any suggestions on how to architect the change?

Thanks!

Matt

kyrrigle:
So I started looking at the code and [this comment](https://github.com/gitbrent/PptxGenJS/blob/8e882ce45e0ea388d42564278a29116e9ed1fb7a/dist/pptxgen.js#L340) seems to suggest that the intention was to rows for labels.

When I put that data into a test and open the Excel sheet I see the opposite.

_ | Jan-17 | Feb-17 | Mar-17 | Apr-17 | May-17
-- | -- | -- | -- | -- | -- 
Red | 11 | 13  | 14 | 15 | 16
[...]

PR #149, opened 2017-08-16
changes needed to support dates in the category axis

kyrrigle:
sample-chart

Changes necessary to support numeric(date formatted) category axis-
- reversed the rows/columns of the data [per this issue](https://github.com/gitbrent/PptxGenJS/issues/145) to support a large number of data values
- added a SharedStrings object to help keep track of the shared strings without using math
- when creating the xlsx check for types and add numbers as numbers
- set the ppt workbook date format explicitly to 1900 based.  (VERY IMPORTANT)
- added an option `catAxisLabelFormatCode` to specify the number format to use for the category labels.  This implies that there should be numbers in the category values. 
- added a new test slide to demo both the date categories and null values

And some others I thought prudent-
- removed the rot="0" on both axis labels.  when rot is not specified, powerpoint will figure out what looks best
- near as I can tell tables/table1.xml isn't used so I've commented it out
- allow for null values with some strategic if statements

As you'll see if you have a look at the sample code you need to supply the dates already in Microsoft 1900 date format.  To do this with momentjs:

```
var nineteen_hundred = moment('1900-01');
var d = moment('2017-07-23T12:35:00Z');
var d1900 = d.diff(nineteen_hundred, 'days', true)+2;
```
(the plus 2 is discussed [here](http://www.cpearson.com/excel/datetime.htm))

and as a bonus if anyone finds themselves wanting to dig through the xml this script may come in handy- [office-unpack.sh](https://gist.github.com/kyrrigle/dbaf3b8b6ce5bdf9c3f1360c7e0b652c)


kyrrigle:
Ok.  I'm done pushing changes to this after the catAxisMajorTimeUnit.   Sorry for pushing it early.

gitbrent:
So, the flipping of the chart axis' has introduced the problem I recalled in Issue #145 - Pie Charts are not editable now.  Select one from the demo slides and click "Edit Data", change one value and the thing becomes corrupted...

gitbrent:
Thanks @kyrrigle !

Features added.  I left out the column flipping as it made data un-editable in Pie Charts.  Fell free to add another PR if we ever get that squared away.

screen shot 2017-09-17 at 23 14 00

PR #163, opened 2017-08-25
Multi Type Charts

clubajax:
Allow for the mixing of two or more chart types. All library chart types are supported. Note that line charts cannot be rotated to be used horizontally (noted because that was a failed goal, due to Powerpoint limitations).
See the readme and demo for docs and examples.

This also adds two options to allow for "DOT" charts (lines with markers-only):
`lineDataSymbolLineSize`: line size of marker
`lineDataSymbolLineColor` color of marker line
`lineSize` is allowed to be `0` for "no line"

The code needed to be refactored a bit - making a chart and the axes now can happen multiple times, so these were made into functions instead of being procedural. Some XML items were re-ordered based on PowerPoint needs and results.

The demo page chart section also was refactored a bit. Each slide is its own function, and they are all called at the end of the section. This allows for easy turning on and off of slides. It's easier to inspect a PPT file with one chart than with dozens.


gitbrent:
Integration work is here: [PR163](https://github.com/gitbrent/PptxGenJS/tree/pr/163)

PR #164, opened 2017-08-25
Demo page update

clubajax:
Made minor change to the demo HTML page to persist the tab selection across refreshes. This is a critical feature since this page doubles as the unit tests. On small screens all the button clicking gets a bit painful.

Also tweaked the tabs CSS to be more responsive an not collapse.


gitbrent:
Hi @clubajax ,

The library needs to retain IE11 compatibility for now, so can you use a non-flex layout?

clubajax:
IE11 supports flexbox. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes

gitbrent:
Thanks @clubajax !

I ended-up siwtching to Flex on the individual tab fieldsets as well, so the whole page display nicely on any size screen.

PR #171, opened 2017-09-05
Escape XML reserved characters so that we don't create invalid xml

kyrrigle:
Added an `escapeXmlEntities` function to escape XML reserved characters and use it wherever labels are being added to xml.  Otherwise we end up with invalid XML.


gitbrent:
Thanks @kyrrigle !

Issue #87, opened 2017-05-17
Image Scaling


necouchman:
Is it possible to have images that are inserted into slides maintain their aspect ratio?  I'm trying to insert images into slides, and it seems like many packages default to allowing you to specify either height or width alone and the image is scaled automatically maintaining aspect ratio.  However, with pptxgenjs's default of 1.0 inches for the height and width, this doesn't appear to work.  Is there a way to scale it while maintaining aspect ratio, or do I need to calculate the aspect ratio manually and then scale appropriately?

gitbrent:
Hi @necouchman ,

Thanks for the question.

The facility to calculate image ratios does not presently exist, so you'll have t calculate it yourself for the time being.

necouchman:
Okay, sounds good - I might try my hand at making the change and then do a pull request.  We'll see how much time I have...

gitbrent:
Thanks - I'll take all the help I can get!  :-)

gitbrent:
This feature has landed with Issue #177.

Thanks to @kajda90 for the code.

danielsiwiec:
Hmm @gitbrent this issue should be reopened. Issue #177 does not solve this problem.

estani:
This hurts... I'm trying to stream as much as possible, so the plan was to stream the files to a dataUri and embedded them (hopping the pptx generation would be streamed in the future as well).
But having to know the size beforehand implies holding the image in memory, as well as the base64 representation... (right?)
In any case this should not be closed, as it's neither implemented or there's a workaround (being the problem, not knowing the image size beforehand)

gitbrent:
Hi @estani 

The upcoming `3.0` release does support streaming.

Can you provide a test case showing image scaling not working? It was my understanding that it had been fixed.

estani:
Sure thing. 
First, thanks for the work and already provided tests, which helped a lot in creating this one.

https://jsfiddle.net/b3t25owj/

As you'll see the original image size is not computed anywhere, it's expected to be provided. Sizing only seems to preserve the "already provided" ratio, so if no `w` and `h` values where provided, they default to 1 so the ratio also defaults to 1. That ratio would be preserved, not the real one.

I provided quite a few slides with tests so perhaps you'll see an error on my part or the documentation. Afaict there's no other options to be set to activate auto-sizing, but maybe I'm wrong.

PR #177, opened 2017-09-07
Image sizing

kajda90:
This feature enables resizing and cropping images. There are three options available – `cover` and `contain` that work the same as CSS property `background-size` and `crop` option that simply cuts off a part of an image (details in readme section _Image Sizing_). 
When this transformation option is used, the original image size (specified by `w` and `h` image properties) still stays unchanged, only image viewport changes. Thus, user can discard or adjust the transformation in PowerPoint.

Note: This PR **does not** solve issue #87.


gitbrent:
Thanks @kajda90 !

I created a new demo Image Slide to showcase the new `sizing` option.

PR #178, opened 2017-09-08
Color validation

kajda90:
This enhancement prevents silent fallbacking invalid color values to black. Now, if the passed color value does not match hexa RGB format or is not one of the listed scheme colors, a warning appears in console.
The fallback stays the same (black color), however developer is notified of passing invalid value.


gitbrent:
Thanks @kajda90 !

PR #184, opened 2017-09-13
Charts: catAxisLabelFrequency property added

kajda90:
Currently, only axis ticks frequency can be controlled. This PR adds options to set even its labels frequency independently on ticks.


gitbrent:
Thanks @kajda90 !

Issue #188, opened 2017-09-15
Promise Dependency - TypeError: Promise.all is not a function


bartolomeu:
When I use **pptx.save()** throws 
> TypeError: Promise.all is not a function

That's caused by this call in the [dist/pptxgen.js:1522](https://github.com/gitbrent/PptxGenJS/blob/master/dist/pptxgen.js#L1522) 
`Promise.all( arrChartPromises )`

in the file [bower.json:10](https://github.com/gitbrent/PptxGenJS/blob/master/bower.json#L10) there is the dependency  of **promise**:
`     "promise": "^0.2.2"`

PromiseJS doesn't have implemented `all()`

The main question is **Promise** is implemented by **ECMAScript 2015 (6th Edition, ECMA-262)** according with [MSN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)
>Chrome - 32.0 
Edge - (Yes)
Firefox - 29.0
Internet Explorer - No
Opera - 19
Safari - 7.1

Just removing the dependency of **promise** will make the the browser use the native Promise from ECMA and finish this problem, that I belive only occours installing through bower.

gitbrent:
Hi @bartolomeu ,

Thanks for letting me know about this.  As you pointed out the only reason to have the polyfill is for IE11, everyone else has it.

Therefore, i've just removed that dependency altogether.  We'll be dropping IE11 support soon anyway, so hopefully there are not many users impacted.

PR #190, opened 2017-09-15
Issue #188 Remove dependency of PromiseJS 

bartolomeu:
Issue #188 Removed dependency of PromiseJS, this will force the browser to use native Promise (from ECMA 2015)


gitbrent:
Thanks @bartolomeu .

The dependency has been removed.

PR #204, opened 2017-09-28
Entities bug

clubajax:
A small patch to fix the missing functionality of using HTML characters in the region property.


gitbrent:
Thanks @clubajax !

PR #205, opened 2017-09-28
Fix axis bug

clubajax:
The catAxis was pointing to the wrong crossing axis. This caused a subtle bug where the 100% grid line would not show up.

This bug/fix can be seen in the charts demo, Slide 5, upper right.


gitbrent:
Thanks @clubajax !

PR #206, opened 2017-09-28
Line chart series colors were not being respected

kyrrigle:
Hi Brent,

Ran into a problem trying to get multi line charts to work.  The attached example ([multi.zip](https://github.com/gitbrent/PptxGenJS/files/1341981/multi.zip)) starts with the docs example but I noticed that I kept getting a purple line for my secondary axis:

image

with this change I can get the black line that I was looking for:

image


gitbrent:
Thanks @kyrrigle !

I completely missed adding the wrapper `<ln>` tag even though it was right above in the `if`... :-)

PR #210, opened 2017-10-06
Fix bar color regression

clubajax:
There was a recent change for _line colors to be respected_ - which overwrote the bar colors being respected.  This is a fix for that.

Also, the `catAxisLabelPos` was removed, which sets the label position for tornado charts. I had documented incorrectly as `tickLblPos`, so it was messed up anyway.


pankiljoshi:
Anyone can answwr please, How to set bar-color?

PR #211, opened 2017-10-07
Pr/210

gitbrent:


gitbrent:
Thanks @clubajax !

Great catch.  The master slide/layout merge and a couple others were hundreds of lines of tangled code and this lost parameter was a victim.

I updated the barchart demo page as well to better showcase the colors and axis label formatting.

Issue #231, opened 2017-11-08
*AxisLineShow options do not work


mconlin:
It appears that both the `catAxisLineShow` and `valAxisLineShow` options don't change anything when generating a chart.  True or false, the axis lines will always show up.

Looking at the source code for building the axes, I couldn't find any reference to these options.  Maybe I'm missing something, but it seems like functionality might have been lost somewhere along the line with regard to these options.

There are some working options that produce a result that is close to what is desired. 
 `catAxisHidden`, for example, does indeed hide the axis line, but it also hides the category text at the same time, which is not desired.

mconlin:
I've created a pull request with a potential fix: https://github.com/gitbrent/PptxGenJS/pull/232

gitbrent:
Thanks @mconlin! The Issue has been fixed in the current codebase.

These options were indeed lost in a past merge.

PR #232, opened 2017-11-08
Fix for Issue #231

mconlin:
This change should make `false` values for `catAxisLineShow` and `valAxisLineShow` properly hide the axis lines.  I attempted to match the already-existing documented functionality for these options, so no change should be needed to the README.


gitbrent:
Thanks @mconlin !

Issue #233, opened 2017-11-08
New Chart option: Legend Text Color


mconlin:
Currently, there appears to be no way to specify a custom color for legend text.  It appears to always default to black.

Other similar aspects of graphs, like titles (`titleColor`) and category labels (`catAxisLabelColor `), can have colors specified.  It would be nice to be able to specify legend colors as well.

mconlin:
I've created a pull request with a potential solution: https://github.com/gitbrent/PptxGenJS/pull/234

gitbrent:
Thanks for the contribution @mconlin !

PR #234, opened 2017-11-08
Update for Issue #233

mconlin:
I added in a new option, `legendColor`.  It behaves similarly to other color options in that it takes a hex color value.  If nothing is specified, PowerPoint seems to always default the color to black.


gitbrent:
Thanks @mconlin !

Issue #240, opened 2017-11-15
Values of 0 (zero) in series are missing in line chart


andrei-cs:
For a series having 0's as values the points don't show up in the graph when viewing the PPT in Powerpoint.

<img width="637" alt="screen shot 2017-11-15 at 15 29 42" src="https://user-images.githubusercontent.com/3409418/32838520-d85687a4-ca19-11e7-98a7-48892a40fd4b.png">

If instead of providing 0's as numerical values I provide them as strings, the chart displays properly.

Code snippets from the provided JSFiddle.

Ex: Not working Values array:

`var pptx = new PptxGenJS();
	var slide = pptx.addNewSlide();
	var opts = { x:1.0, y:1.0, font_size:42, color:'00FF00' };
	slide.addText('Hello World!', opts);
var chartData = [
  {
    "name": "Chart",
    "labels": [
      "16-Oct-17",
      "17-Oct-17",
      "18-Oct-17",
      "19-Oct-17",
      "20-Oct-17",
      "21-Oct-17",
      "22-Oct-17",
      "23-Oct-17",
      "24-Oct-17",
      "25-Oct-17",
      "26-Oct-17",
      "27-Oct-17",
      "28-Oct-17",
      "29-Oct-17",
      "30-Oct-17",
      "31-Oct-17",
      "01-Nov-17",
      "02-Nov-17",
      "03-Nov-17",
      "04-Nov-17",
      "05-Nov-17",
      "06-Nov-17",
      "07-Nov-17",
      "08-Nov-17",
      "09-Nov-17",
      "10-Nov-17",
      "11-Nov-17",
      "12-Nov-17",
      "13-Nov-17",
      "14-Nov-17",
      "15-Nov-17"
    ],
    "values": [
      0,
      0,
      149,
      141,
      0,
      0,
      0,
      0,
      0,
      0,
      139,
      1177,
      477,
      0,
      0,
      2578,
      0,
      0,
      349,
      0,
      0,
      0,
      2195,
      1057,
      807,
      0,
      0,
      0,
      148,
      241,
      0
    ]
  }
]
slide.addChart( pptx.charts.LINE, chartData)
pptx.save();`

Ex: Working Values Array

`var pptx = new PptxGenJS();
	var slide = pptx.addNewSlide();
	var opts = { x:1.0, y:1.0, font_size:42, color:'00FF00' };
	slide.addText('Hello World!', opts);
  var chartData = [
  {
    "name": "Chart",
[...]

PR #243, opened 2017-11-19
Fix error that causes 0 values to be the empty string

hilnius:
the `value || ""` replacement was making 0 values replaced by '' (the empty string) instead of keeping 0 which is a valid value.


gitbrent:
NOTE: This fix is for Issue #240

gitbrent:
Thanks @hilnius !

hilnius:
@gitbrent No problem, thanks for all the rest :)

PR #248, opened 2017-11-22
Issue 83 multiple presentations in node

labriola:
Effectively we expose an additional property, in the case where the user is using NodeJs on the public API. The additional returned property is the constructor, hence a node user can now destructure the returned object and make multiple presentations:

const {
  PptxGenJS 
} = require('pptxgenjs');

const pptx1 = new PptxGenJS();
const pptx2 = new PptxGenJS();

This won't impact backward compatibility from a functional perspective. The only noticeable side effect would be for users currently requiring the entire public API. They would now see an additional method present on the API. That is a compromise but it would solve the core issue and bring the project in line with how node recommends using libraries.

Then it can be fixed better/differently when a breaking change is appropriate. Let me know if this works for you.


gitbrent:
Hi @labriola,

Thanks very much for the patch.

I've decided to go ahead and implement Node module exports in the standard fashion and 2.0 is in beta now.

PR #309, opened 2018-03-14
Adding rounded corners to images (avatar...)

hoangpq:
@gitbrent I faced a problem when try to show up the rounded avatar without rounding first so i created that PR to fix that problem in some cases we need to show rounded corners images. 

image

Image options will allow to pass `rounding` attribute like 
```javascript
slide.addImage({
  x: 1.0, y: 1.0, w: 8.0, h: 4.0,
  hyperlink: { url: 'https://github.com/gitbrent/pptxgenjs', tooltip: 'Avatar' },
  path: 'images/monkey.jpg',
  rounding: true // default: false 
});
```


gitbrent:
Thanks @hoangpq !

I've added rounding to the Image demo slide.

screen shot 2018-03-19 at 22 31 03

PR #340, opened 2018-05-22
allow passing custom 'description' for images for 508 compliance

henrylavoo:


NETEHenry:
Sorry, meant to only do this PR to my fork. Doesn't look like I can rescind this PR. I'm going to be working on a few updates for 508 compliance and may do a PR in the future once I've added more features.

PR #342, opened 2018-05-31
Fixed Error thrown while rendering Area Chart

KrishnaTejaReddyV:
The edited line in makeChartType() function initially compared opts.type to a string while it is an object containing a property name. Hence it must be changed to :

opts.type.name != 'area'


gitbrent:
Thanks @KrishnaTejaReddyV !

PR #344, opened 2018-06-01
Fixed Title display on showTitle = false error

KrishnaTejaReddyV:
In Microsoft PowerPoint 2007, even the absence of <title> tag in xml on setting showTitle=false is creating a title by default with a value of 'name' key of the JSON input. This can be fixed by setting <c:autoTitleDeleted val="1"/> tag when rel.opts.showTitle is false. 

This problem does not exist in the later versions of PowerPoint but is present in PowerPoint 2007 and the fix works at all places.


gitbrent:
Thanks @KrishnaTejaReddyV !

PR #365, opened 2018-06-26
Added option displayBlankAs

### LINKED ISSUE #364: Blanks display in line charts
Hi Guys ! 

First, thank you very much for your work on this library, it's been very useful to us where I work, we use it every day to generate presentations and It works very well 😉

We had an issue while generating line graphs, some value where null and so there was a gap between values like this :
capture d ecran 2018-06-26 a 15 39 17

We wanted the line to continue between the gap (it can be done directly in powerpoint) but I didn't find any option in pptxgenjs to change this behaviour. My colleague saw by changing that option in powerpoint and making a diff of the two files that this had to do with one xml tag (`<c:dispBlanksAs val="gap"/>`). If it is replaced with : `<c:dispBlanksAs val="span"/>`, it works :
capture d ecran 2018-06-26 a 16 40 15

Would you guys be ok with adding an option `displayBlankAs` in the graphs options ? I'm making a PR right now 😉

guipas:
closes #364 
As promised ;-)

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -2581,7 +2581,7 @@ var PptxGenJS = function(){
 		}
 
 		strXml += '  <c:plotVisOnly val="1"/>';
-		strXml += '  <c:dispBlanksAs val="gap"/>';
+		strXml += '  <c:dispBlanksAs val="' + (rel.opts.displayBlankAs ? rel.opts.displayBlankAs : 'gap') + '"/>';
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks @guipas !

gitbrent:
FYI: I changed the name to `displayBlanksAs` (blanks, plural) to match the actual XML tag, and set the default to be 'span' as it looks much better.

[Chart Docs: Data Options](https://gitbrent.github.io/PptxGenJS/docs/api-charts.html#chart-data-options)

screen shot 2018-06-27 at 23 21 34

PR #367, opened 2018-06-28
New Feature - add ability to hide slides

ReimaFrgos:
Add the ability to set a specific slide as hidden. Default state is that a slide is visible.

Usage:
`slide.hideSlide = true;`


gitbrent:
Thanks @ReimaFrgos !

I renamed the property to `hidden`.

### [Slide Options](https://gitbrent.github.io/PptxGenJS/docs/usage-slide-options.html#slide-hiding)

PR #371, opened 2018-07-04
Removed error thrown while rendering Multi Type chart containing Area

KrishnaTejaReddyV:
Data label position is being set to outEnd for Multitype charts that contain an area chart as the other chart type would enable inclusion of that line. Hence either the suggested changes can be made or the entire line checking the condition can be removed.


gitbrent:
Hi @KrishnaTejaReddyV ,

Thanks for the submission.

Can you provide some code/screencaps showing the issue before and after the fix?

KrishnaTejaReddyV:
img-20180705-wa0008
img-20180705-wa0005
img-20180705-wa0006

Attached the images with error before change, output before change and output after change

gitbrent:
Thanks @KrishnaTejaReddyV !

PR #372, opened 2018-07-06
Irregular Display of Cat Axis for Scatter and Bubble

KrishnaTejaReddyV:
In case of Scatter and Bubble charts, the category axis also contains numbers just like value axis. Therefore, on putting values having more than 2 digits in cat axis, the display of category labels is being compromised. Therefore the charts should contain two Val axises in case of Scatter and Bubble.

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -3297,7 +3297,13 @@ var PptxGenJS = function(){
 		var strXml = '';
 
 		// Build cat axis tag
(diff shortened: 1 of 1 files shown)


KrishnaTejaReddyV:
img-20180706-wa0013
img-20180706-wa0012

gitbrent:
Thanks @KrishnaTejaReddyV !

PR #415, opened 2018-10-06
Added better NodeJS checking and better support for Angular

antiremy:
I know this works for my cases, but I'd like someone else who's using PptxGenJS in Angular to test this out.


gitbrent:
Thanks @antiremy !

I was not familiar enough with Angular to understand to test for `typeof window !== 'undefined'`.

I hope this solves a lot of people's issues.

PR #418, opened 2018-10-12
Replace jquery each by standard forEach

fdussert:
I don't know why but this $.each don't work in a node/react app.
It triggers an error and prevent the file to be generated and saved


gitbrent:
@fdussert - thanks for reporting this and creating a pull request.

kylesimmonds:
@gitbrent any idea when this is scheduled to be published to npm?

gitbrent:
Probably by month's end.  SVG support is still needed for non-browsers, otherwise, `2.4.0` is ready.

Issue #410, opened 2018-10-01
[Feature Request] Adding label to Scatter plot


nevermore2014:
Hello @gitbrent !

I really appreciate the library that you created, I happen to have a small feature not sure if is possible? So for generating a XY SCATTER chart would it be possible to add label to each of the data point? Right now you can only add label to Bar chart or Radar chart, not Scatter chart.

// Chart Type: XY SCATTER var dataChartScatter = [ { name:'X-Axis', values:[1,2,3,4,5,6,7,8,9,10] }, { name:'Y-Value 1', values:[13, 20, 21, 25] }, { name:'Y-Value 2', values:[21, 22, 25, 49] } ]; slide.addChart( pptx.charts.SCATTER, dataChartScatter, { x:1.0, y:1.0, w:6, h:4 } );

Let me know, thanks!

Hongyu

gitbrent:
Hi @nevermore2014 ,

If you'd like to look through the code and see what it would take to add this feature via a pull request, i'd encourage you to do so.

nevermore2014:
@ReimaFrgos oh man, who are you.. you are just the equalizer to me, thanks so much!! I would love to buy you a cup of coffee if you are in great Boston area!

jsvishal:
Do we have support for labels in BUBBLE CHART?

Issue #291, opened 2018-02-20
Scattered Chart - Labels for each set of XY


pankajsindagi:
Hi,

Currently the labels for each data point in scattered chart is picked from the `name `of the **Y-Axis**. It would be excellent feature to have individual labels for each data point. Looking forward for this feature in next release.

Thanks Brent :)

gitbrent:
Hi @pankajsindagi ,

Can you provide an example?

pankajsindagi:
When we create ppt slide with below chart type,

```
// Chart Type: XY SCATTER
var dataChartScatter = [
	{ name:'X-Axis',    values:[1,2,3,4,5,6,7,8,9,10] },
	{ name:'Y-Value 1', values:[13, 20, 21, 25] },
	{ name:'Y-Value 2', values:[21, 22, 25, 49] }
];
slide.addChart( pptx.charts.SCATTER, dataChartScatter, { x:1.0, y:1.0, w:6, h:4 } );
```

![image](https://user-images.githubusercontent.com/20287647/36480769-83d8d3e6-1733-11e8-9999-ff468f2182fc.png)

When hover on data point, as you can see in above chart. We see the name as label, i.e `Y-Value 1`. In practical scenarios it would be ideal to have separate label which represents the particular data point i.e (4, 25) 

Please consider this helpful feature.

Again, Thanks for the cool library :)

pankajsindagi:
When we create ppt slide with below chart type,

```
// Chart Type: XY SCATTER
var dataChartScatter = [
	{ name:'X-Axis',    values:[1,2,3,4,5,6,7,8,9,10] },
	{ name:'Y-Value 1', values:[13, 20, 21, 25] },
	{ name:'Y-Value 2', values:[21, 22, 25, 49] }
];
slide.addChart( pptx.charts.SCATTER, dataChartScatter, { x:1.0, y:1.0, w:6, h:4 } );
```

![image](https://user-images.githubusercontent.com/20287647/36480769-83d8d3e6-1733-11e8-9999-ff468f2182fc.png)

When hover on data point, as you can see in above chart. We see the name as label, i.e `Y-Value 1`. In practical scenarios it would be ideal to have separate label which represents the particular data point i.e (4, 25) 

Please consider this helpful feature.

Again, Thanks for the cool library :)

pankajsindagi:
I've data something like this:

```
var data = [ 
                   {label: 'Red 1', x: 1, y:13 },
[...]

PR #420, opened 2018-10-19
New Feature: Add Data Labels to Scatter Charts

ReimaFrgos:
New feature to allow labels to be added on Scatter Charts. Should resolve issues #410 and #291 

The following options are used for labelling scatter charts:
```
showLabel:true // Must be set to true or labels will not be shown
dataLabelFormatScatter: custom // Can be set to custom (default), customXY, or XY. 
```

There are 3 different formats that the label can be set as, and defined using dataLabelScatterFormat:
- custom - Any custom string, can be unique per plot. e.g. 'My Label'
- customXY - Any custom string with the XY plot coordinates appended. e.g. 'My Label (X,Y)'
- XY - Uses the XY plot coordinates as the label. e.g X, Y

For customXY and XY, the plot references are dynamic and will update if the underlying excel data is updated.

Data labels are applied to the Y-Values, and support multiple sets of data in the format.
```
var data = [
    { name:'X-Axis',    values:[10,12,32,47] },
    { name:'Y-Value 1', values:[13, 20, 21, 25], labels:['Red 1', 'Red 2', 'Red 3', 'Red 4'] },
    { name:'Y-Value 2', values:[21, 22, 25, 49], labels:['Blue 1', 'Blue 2', 'Blue 3', 'Blue 4'] }
];
```
You do not need to specify the labels object if using `dataLabelFormatScatter: XY` as only the values are used in the labels.

If using `dataLabelFormatScatter: XY` every plot point will be labelled.

If using `dataLabelFormatScatter: custom` or `dataLabelFormatScatter: customXY` you can choose to label only some plot points by passing in a blank label. Useful when plotting a large number of points e.g. 
```
var data = [
    { name:'X-Axis',    values:[10,12,32,47] },
    { name:'Y-Value 1', values:[13, 20, 21, 25], labels:['Red 1', '', '', 'Red 4'] },
    { name:'Y-Value 2', values:[21, 22, 25, 49], labels:['Blue 1', 'Blue 2', 'Blue 3', ''] }
];
```

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -635,6 +635,9 @@ var PptxGenJS = function(){
 			if ( !options.dataLabelFormatCode && options.type.name === 'scatter' ) options.dataLabelFormatCode = "General";
 			options.dataLabelFormatCode = options.dataLabelFormatCode && typeof options.dataLabelFormatCode === 'string' ? options.dataLabelFormatCode : (options.type.name == 'pie' || options.type.name == 'doughnut') ? '0%' : '#,##0';
 			//
+			// Set default format for Scatter chart labels to custom string if not defined
+			if ( !options.dataLabelFormatScatter && options.type.name === 'scatter') options.dataLabelFormatScatter = 'custom';
+			//
 			options.lineSize = ( typeof options.lineSize === 'number' ? options.lineSize : 2 );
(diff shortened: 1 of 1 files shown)


jsvishal:
Can this be added for Bubble Chart too?

PR #422, opened 2018-10-19
BugFix: 0 value plot points ignored on Scatter Chart

ReimaFrgos:
Fixes a bug on scatter charts where if either the X or Y value was a 0, the data would be ignored and the plot point not shown on the chart.


gitbrent:
Thanks @ReimaFrgos !

PR #424, opened 2018-10-23
Pass the callback as a function, rather than invoke it

### LINKED ISSUE #423: Save callback is not called as a 'callback'
The save callback is immediately invoked, instead of being passed to the file writer as a function.

danielsiwiec:
Closes #423


gitbrent:
Thanks @danielsiwiec !

Note: Fixes issue with `2.3.0`:

Ex:
```javascript
pptx.save( exportName, (filename) => { console.log('inline callback!'); } );
```
Output:
```bash
(node:16957) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
```

PR #431, opened 2018-11-14
remove legacy/deprecated attribute

efx:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes.
`lang` was specified here, so probably ignored by most browsers.


gitbrent:
Thanks @efx!

PR #434, opened 2018-11-19
BugFix: figuring all nums<100 are inches breaks inserting of small images

antonandreyev:
Currently all sized less than "100" are considered to be in inches (see getSmartParseNumber function). This cause a presentation corruption when inserting small images because their dimensions are being converted. For example image with height 75 px is converted to be about 600000000 px high. I propose to make the border to be at least "50" (could be lowered imo) which should not affect real "inch" sizes since Slide width is about 13''.

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -2149,11 +2149,11 @@ var PptxGenJS = function(){
 
 		// CASE 1: Number in inches
 		// Figure any number less than 100 is inches
-		if ( typeof inVal == 'number' && inVal < 100 ) return inch2Emu(inVal);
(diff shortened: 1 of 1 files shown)


gitbrent:
Hi @antonandreyev 

All units should be inches.  If you're passing `75` as a pixel size, that is incorrect.

Please see: `examples\pptxgenjs-demo.html#tab4` and `examples\pptxgenjs-demo.js` 

EX: Image 3 inches wide, 3 inches tall
```
slide.addImage({ path:gPaths.ccCopyRemix.path, x:1, y:1, w:3.0, h:3.0 });
```

PR #435, opened 2018-11-19
BugFix: missing jpg content type breaks corrupts presentation for Office365

antonandreyev:
By default we add only "jpeg" into presentation's content types (see makeXmlContTypes function) but Office365 definitely expects to see "jpg" also and consider some presentations as corrupted depending on certain pictures inserted.


gitbrent:
Thanks @antonandreyev !

I have not experienced problems with images on office.com, so thanks for the information and the patch.

screen shot 2018-12-11 at 20 37 35

PR #436, opened 2018-11-19
BugFix: using $ for jQuery leads to conflict with code of some applications

antonandreyev:
"$" symbol is reserved for usage in other web-applications (e.g. SharePoint for me). So it is better to use more specific synonym "jQuery". For me using $ breaks only pptxgenjs, for others it may break their own apps.


gitbrent:
Thanks @antonandreyev

Issue #433, opened 2018-11-17
addImage not working with parameter in URL


katsuya-horiuchi:
If URL contains parameter, passing `path` to `addImage` function leads to data corruption of file.

Example:
```javascript
var pptx = new PptxGenJS();
var slide pptx.addNewSlide();
slide.addImage({
  path: "https://gitbrent.github.io/PptxGenJS/img/feature-callout.png?key=value",
  x: 1,
  y: 1
})
pptx.save("test");
``` 

Would it be possible to include images whose URL path has parameter?

PR #437, opened 2018-11-20
Remove query and fragment from image's extension (Issue #433)

katsuya-horiuchi:
Not setting extension of image correctly seems to result in data corruption in file.
As far as the issue goes, leaving fragment in URL seems file (at least in LibreOffice), but the change will remove it as well just in case.

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -405,7 +405,7 @@ var PptxGenJS = function(){
 			}
 
 			// STEP 1: Set extension
-			var strImgExtn = strImagePath.split('.').pop() || 'png';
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks @katsuya-horiuchi !

ありがとう

katsuya-horiuchi:
Glad to be of help. However, I found another issue, so I'll file an issue after further investigation.

And nice use of Japanese :)

gitbrent:
Sure, just open the `examples/pptxgenjs-demo.html` file in your browser.

You can test any/all features. :-)

Issue #440, opened 2018-11-27
addImage not working with multiple parameters in URL


katsuya-horiuchi:
Multiple parameters combined with ampersand (&) will lead to data corruption.

Example:
```javascript
var pptx = new PptxGenJS();
var slide pptx.addNewSlide();
slide.addImage({
  path: "https://gitbrent.github.io/PptxGenJS/img/feature-callout.png?key=value&key2=value2",
  x: 1,
  y: 1
})
pptx.save("test");
```

I haven't been able to spare time to look into what the library does under the hood, but I have found one thing that might help solve the issue.
Escaping `&` in `path` (i.e. `https://gitbrent.github.io/PptxGenJS/img/feature-callout.png?key=value&amp;key2=value2`) will solve the issue. However, the request sent to web server will not adjust escaping character, so web server might fail to return appropriate response.

With that said, in order to solve the issue, we need to
1. escape ampersand in parameter appropriately (i.e. `&` to `&amp;`)
2. readjust the param when sending request to web server (`&amp;` to `&`)
3. use escaped character (`&amp;`) when writing url (I'm guessing the library converts data into xml or something).

I'm sure working on 1 is easy, but I don't know if we can work on 2 and 3.

gitbrent:
Hi @katsuya-horiuchi 

The image loads and is included in the PPTX correctly, the issue with the corrupted slide stems from including the full filename as the "Description"

Adding a call to encode the XML in the `descr` field fixes the issue (line:1111)
`descr="'+ encodeXmlEntities(slideItemObj.image) +'"`

![screen shot 2018-12-02 at 11 45 15](https://user-images.githubusercontent.com/7218970/49343005-1bc7be00-f628-11e8-9f87-feffe3694ec6.png)

PR #442, opened 2018-11-29
Escape ampersands in URL parameter (Issue #440)

katsuya-horiuchi:
This is the fix for issue #440.

Upon the investigation, I found out that escaping ampersands in `resultObject` seems to be enough to solve the issue.
The fix itself looks rather complicated, but I also considered cases with multiple `?` in URL.


gitbrent:
Fixed via changing line:1111 to XML encode image `descr` field

PR #443, opened 2018-11-29
Export more types to enhance TypeScript support

ntietz:
While using PptxGenJS with TypeScript, I wanted to do something like this:
```
function addChartToSlide(slide: Slide, chartData: ChartData): void {
  ...
}
```

With the existing type definitions, type inference works as expected but the Slide type (among others) are not exported and thus are not able to be referenced from client code, making it more difficult to write correct type definitions.

This PR exports all the types defined within the PptxGenJS namespace, while retaining the current behavior (export the PptxGenJS *class* as the default export).


ntietz:
@gitbrent Good catch. I didn't notice that as I had a problem in my local workflow. This is resolved, and I've verified that no problems were detected locally - for my machine. Thanks for taking the time to check out this PR :smile: 

I've updated the diff to be smaller, based on [a template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html) from the TypeScript Handbook. I think this is cleaner, and results in the same end result: now Slide, CommonOptions, etc. can all be referenced as well to allow more typechecking in client code.

gitbrent:
Great work and thanks for the contribution @ntietz !

ntietz:
Thanks for creating and maintaining this and being responsive!

PR #474, opened 2019-01-14
New Chart Option: valAxisCrossesAt

ReimaFrgos:
Adds a new chart option: valAxisCrossesAt

This allows you to specify where the Category axis crosses the Value Axis.

Acceptable values are:
1. autoZero - Default if not entered
2. max
3. A number within the range of the Value Axis min and max value


gitbrent:
Thanks @ReimaFrgos !

cpf121:
Do you have a similar catAxisCrossesAt option?

PR #509, opened 2019-03-01
Link directly to the StackOverflow PptxGenJS tag

SKalt:
This links (1) to prior answers tagged `PptxGenJS` and (2) to a new question template including the `PptxGenJs` tag.


gitbrent:
Thanks @SKalt !

Issue #464, opened 2018-12-27
Extending Master Slide Object Shapes Using React/Webpack


cavaloni:
Currently, it appears that `MASTER_OBJECTS` is not extensible to those objects/shapes available in `pptxgen.shapes.js`. If I am understanding your code well enough, would it seem reasonable in your design to be able to pass `gObjPptxShapes` into the instantiation of `PptxGenJS` and conditionally extend the object `MASTER_OBJECTS` to include them if they are present? (also adding appropriate calls to `addShapeDefinition`).

I am using React and so importing the script with `<script>` is not desirable so it would also help on that level. But primarily, I would like to be able to create a master slide with more shapes than what seems to be readily available. If this seems like a reasonable approach then I would be happy to submit a PR.

cavaloni:
Should I just go ahead and submit a PR so you can see what I'm referring to @gitbrent?

gitbrent:
Hi @cavaloni 

Sure, let me see what you're thinking of as a solution here.

cavaloni:
Ok @gitbrent, here's what I've got. Ready to submit a PR if you're good with the idea.

Very simple change, but you can let me know if it is in keeping with your patterns enough or not, and I can change as needed.

Starts with passing in `gObjPptxShapes` when instantiating a `new PptxGenJS` instance, and checks for existence here and assigns :

```
var PptxGenJS = function(shapesExt){
	if (!gObjPptxShapes && !!shapesExt) gObjPptxShapes = shapesExt;
```

Then it extends `MASTER_OBJECTS` if it extists:

```
if (gObjPptxShapes) {
		Object.keys(gObjPptxShapes).forEach(function(shapeName) {
			// FIXME If needed, to alter name best for the api
			const masterObjectName = gObjPptxShapes[shapeName].displayName.toLowerCase();
			MASTER_OBJECTS[masterObjectName] = { name: masterObjectName, objName: shapeName };
		})
	}
```

Then the check for `MASTER_OBJECTS` looks like so (primary changes are in the beginning, checking for existence on the object, then the last `else if` statement which adds it):

```
slideDef.objects.forEach(function(object,idx){
					var key = Object.keys(object)[0];
					var masterObject = MASTER_OBJECTS[key];
					
					if (!MASTER_OBJECTS[key]) return;

[...]

PR #512, opened 2019-03-07
Extend mstr_objcts for use in mst slides.

cavaloni:
Reference issue #464 for extending the master slides ability to use more shape objects.


gitbrent:
Hi @cavaloni 

This should be resolved in the upcoming `3.0` release as it's TypeScript now and built with rollup, correctly exposing these vars.

PR #528, opened 2019-03-31
fixing the SVG support

RicardoNiepel:
There are currently multiple bugs inside the code base for the SVG support:
- the PNG DataURL is also assigned to the SVG version
  - written SVG file in the media folder isn't a SVG, it's the PNG - thus Powerpoint cannot load it correctly
- the mime type for SVG is wrong

This PR fixes all - now the real SVG is used inside the pptx.
You can easily test it with 'pptxgenjs-demo.html': just scale the SVGs before and after this PR.


mmayer1:
This fix works great for me! Looking forward to having a new release with this merged

gitbrent:
Thanks @RicardoNiepel !

Issue is fixed in Version 3.0

### Previous version:
Screen Shot 2019-09-23 at 23 02 09

### Version 3.0:
Screen Shot 2019-09-23 at 23 02 01

Issue #414, opened 2018-10-04
outputType is ignored when using elements with path that need to be converted to base64


sijmenvos:
If you add images with a path, when saving the presentation it ignores the chosen outputType.

Simple code to reproduce:
```javascript
const PptxGenJS = require('pptxgenjs');

const pptx = new PptxGenJS();
const slide = pptx.addNewSlide();
slide.addImage({
  path: 'https://via.placeholder.com/350x150',
  w: '100%',
  h: '100%',
});

pptx.save('jszip', (result) => {
  console.log(result); // === 'jszip.pptx' file saved locally instead of base64 encoded string
}, 'base64');
```

After converting all images to base64, the outputType isn't passed to the doExportPresentation function:
https://github.com/gitbrent/PptxGenJS/blob/9f23695edd96752052d3fff1ba70ed373d8b5b57/dist/pptxgen.js#L2136

sijmenvos:
Found a workaround if you want to export type to a nodebuffer; if the export name starts with 'http', it will return as 'nodebuffer' type:

```javascript
const PptxGenJS = require('pptxgenjs');

const pptx = new PptxGenJS();
const slide = pptx.addNewSlide();
slide.addImage({
  path: 'https://via.placeholder.com/350x150',
  w: '100%',
  h: '100%',
});

pptx.save('http', (result) => {
  console.log(result); // === Buffer<>
}, 'nodebuffer');
```

gitbrent:
Hi @sijmenvos 

This is fixed in version `3.0`

PR #531, opened 2019-04-17
Fix Issue 414: outputType being Ignored

hktony0129:
Bug fix for issue #414 

Previously, `outputType` was not passed to `doExportPresentation()` when the function gets called in `callbackImgToDataURLDone()`. This mean that if the output pptx had any data that required base64 conversion, the output type would be disregarded.

This PR sets `outputType` to be part of `gObjPptx` when the user calls `pptx.save()`. Then when `doExportPresentation()` gets called, `outputType` will be fetched from `gObjPptx`

### DIFF
--- a/dist/pptxgen.js
+++ b/dist/pptxgen.js
@@ -1761,7 +1761,7 @@ var PptxGenJS = function(){
 	/**
 	 * DESC: Export the .pptx file
 	 */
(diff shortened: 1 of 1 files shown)


gitbrent:
Hi @hktony0129 

This is fixed in version `3.0`

PR #550, opened 2019-06-03
Correct and improve TS type definitions

iota-pi:
Corrected, updated and improved TypeScript type definition file

### DIFF
--- a/dist/pptxgen.d.ts
+++ b/dist/pptxgen.d.ts
@@ -1,10 +1,10 @@
-// Type definitions for pptxgenjs 2.3.0
+// Type definitions for pptxgenjs 2.5.0
 // Project: https://gitbrent.github.io/PptxGenJS/
 // Definitions by: Brent Ely <https://github.com/gitbrent/>
 //                 Michael Beaumont <https://github.com/michaelbeaumont>
 //                 Nicholas Tietz-Sokolsky <https://github.com/ntietz>
-// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+//                 David Adams <https://github.com/iota-pi>
+// TypeScript Version: 3.x
 
 export as namespace PptxGenJS;
 
@@ -63,7 +63,7 @@ declare namespace PptxGenJS {
     h?: Coord;
   }
 
-  export type CommonOptions = PositionOptions; // for backwards compatability
+  export type CommonOptions = PositionOptions; // for backwards compatibility
 
   export type Hyperlink = ({ url: string; slide?: undefined } | { slide: number; url?: undefined }) & { tooltip?: string };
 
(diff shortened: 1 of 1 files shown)


dist/pptxgen.d.ts
```diff
@@ -34,67 +34,325 @@ declare class PptxGenJS {
...
-  export interface DataOrPath {
-    // Exactly one must be set
-    data?: string;
-    path?: string;
-  }
-  export interface ImageOptions extends CommonOptions, DataOrPath {
+
+  export type DataOrPath = { data: string; path?: undefined } | { data?: undefined; path: string };
+
+  export interface ImageBaseOptions extends CommonOptions {
```
adrienarcuri: Hello @iota-pi 
Thank you for improving TS definition.

Could you insert Hyperlink and ImageSizingOptions in your Pull Request please ?
`
  export interface ImageBaseOptions extends CommonOptions {
    hyperlink?: Hyperlink;
    rounding?: boolean;
    sizing?: ImageSizingOptions;
  }

  export interface ImageSizingOptions extends CommonOptions {
    type?: 'cover' | 'contain' | 'crop';
  }

  export type Hyperlink = ({ url: string; slide?: undefined } | { slide: number; url?: undefined }) & { tooltip?: string };
`

dist/pptxgen.d.ts
```diff
@@ -34,67 +34,325 @@ declare class PptxGenJS {
...
+    fontSize?: number;
+  }
+
+  export interface TextOptions extends ShapeOptions, BasicTextFormatting {
+    align?: HAlign;
+    autoFit?: boolean;
+    breakLine?: boolean;
+    bullet?: boolean | { type: 'number'; code?: undefined } | { code: string; type?: undefined };
+    charSpacing?: number;
+    hyperlink?: ({ url: string; slide?: undefined } | { slide: number; url?: undefined }) & { tooltip?: string };
```
adrienarcuri: We can use Hyperlink type:
` hyperlink?: Hyperlink;`

iota-pi:
Also renamed `CommonOptions` to `PositionOptions` for clarity, but added alias for backwards compatibility

gitbrent:
Thanks @iota-pi and @adrienarcuri !

thitemple:
Hey there, quick question, do you thing it would be helpful to make the shape names a string literal? Something like this:

```
export type ShapeType = "LINE" | "RECTANGLE" | "OVAL" | ...;
```

I'd be glad to add that

iota-pi:
There are rather a lot of them... ideally the `pptxgen.shapes.js` file would be a JSON or TS file and we could just do
```
export type ShapeType = keyof gObjPptxShapes;
```

thitemple:
We can't do that today because that file is a JS file. Turning it into a JSON or TS file would add even more changes and I'm not sure if that's the author wants if that's the case I could work on that.

If not, I have that literal type defined already, I could just add it to the typings file.

gitbrent:
Actually, the author has rewritten the library in TS already (the 3.0 branch) and would appreciate any help with definitions, etc. :-)

PR #558, opened 2019-06-12
Adds typings reference to package.json

thitemple:
So that TypeScript projects can actually use the existing typings defined in the project.


gitbrent:
Thanks @thitemple !

PR #560, opened 2019-06-13
Keep output type when containing images

baidi:


gitbrent:
Thanks @baidi 

This is fixed in 3.0

PR #614, opened 2019-10-30
NodeJS, and download bug fixes

JeffSkinn:
fixed issue with Youtube link requires link parameter, fixed issue with nodeJS always returning false.


gitbrent:
Hi @JeffSkinn 

Thanks for the contribution - both of these are fixed in the 3.0 branch.

PR #620, opened 2019-11-27
Text in slides does not override the bullet master

sgenoud:
In the previous version, it overrides it by default

### DIFF
--- a/src/gen-xml.ts
+++ b/src/gen-xml.ts
@@ -911,8 +911,10 @@ function genXmlParagraphProperties(textObj: ISlideObject | IText, isDefault: boo
 				bulletLvl0Margin +
 				'"'
 			strXmlBullet = '<a:buSzPct val="100000"/><a:buChar char="' + BULLET_TYPES['DEFAULT'] + '"/>'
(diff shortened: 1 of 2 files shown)


src/gen-xml.ts
```diff
@@ -911,8 +911,10 @@ function genXmlParagraphProperties(textObj: ISlideObject | IText, isDefault: boo
 				bulletLvl0Margin +
 				'"'
 			strXmlBullet = '<a:buSzPct val="100000"/><a:buChar char="' + BULLET_TYPES['DEFAULT'] + '"/>'
-		} else {
-			strXmlBullet = '<a:buNone/>'
+		} else if ( textObj.options.bullet === false ){
+        // We only add this when the user explicitely asks for no bullet.
+        // Otherwise it can override the master defaults
```
ankon, line 916: The spaces in the `if` don't match the existing style (space before '(', space after ')', no space after '(', no space before ')'), and the comment is indented in an odd way.

gitbrent:
This merged into an old branch, so i've added the code myself.

PR #621, opened 2019-11-27
The paragraph body of placeholder is formatted

sgenoud:
This was breaking anchoring (valign) for placeholder.

While I was there, I also added support for `valign = middle` that was broken (while center worked)


kevinresol:
Just realize I have also submitted PR for this, but separated: #632 and #631

PR #622, opened 2019-11-27
DataTable font size option

MehdiAroui:
Adding the ability to set the font size for Charts DataTable


gitbrent:
Thanks @MehdiAroui !

PR #631, opened 2019-12-30
Fix valign handling, it should support "middle" according to the doc

kevinresol:


gitbrent:
Thanks @kevinresol !

Code fixed by changing the `'c'` string to `'m'`
The type remains 'ctr' as thats the valid XML - "middle" is used as the valign option to match PowerPoint UI.

PR #633, opened 2019-12-30
Round margin values

kevinresol:
because PowerPoint will consider floating point values in the xml file as invalid value


gitbrent:
Thanks @kevinresol !

PR #639, opened 2020-01-05
Add ability to rotate image

alabaki:
It is now possible to add a 'rotate' value to the addImage() function

### DIFF
--- a/src/core-interfaces.ts
+++ b/src/core-interfaces.ts
@@ -207,6 +207,7 @@ export interface IImageOpts extends PositionOptions, OptsDataOrPath {
 	hyperlink?: HyperLink
 	rounding?: boolean
 	placeholder?: any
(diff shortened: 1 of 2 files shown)


src/core-interfaces.ts
```diff
@@ -207,6 +207,7 @@ export interface IImageOpts extends PositionOptions, OptsDataOrPath {
 	hyperlink?: HyperLink
 	rounding?: boolean
 	placeholder?: any
+	rotate?: number
```
cronin4392: Does this also need to be added to `types/index.d.ts` @gitbrent?

gitbrent:
Thanks @alabaki !

src/core-interfaces.ts
gitbrent: Yes, good catch!

PR #641, opened 2020-01-08
Include types in package.json files

cronin4392:
When I cloned this repo down there were no types included since they are no longer in the `dist` directory. This will include them in the downloaded package files.


gitbrent:
Ack!! I didn't even spot this...

Thanks @cronin4392

PR #643, opened 2020-01-09
Add missing margin type from ITextOpts

cronin4392:
This type is listed in the documentation and does work, but was not in the types file. I think these are the only two places it's needed to be added?


gitbrent:
Thanks @cronin4392 !

PR #650, opened 2020-01-14
Add shapes and font options types

cronin4392:


src/core-interfaces.ts
```diff
@@ -23,6 +23,10 @@ export type ShapeFill = Color | { type: string; color: Color; alpha?: number }
 export type BkgdOpts = { src?: string; path?: string; data?: string }
```
cronin4392: @gitbrent would you mind explaining your thinking behind having both `core-*.ts` files as well as `types/index.d.ts`? My assumption is that a goal would be to get rid of one of these files and have types coming from one place?

cronin4392: I dug in a bit more and noticed that `tsconfig.js` has `declarationDir` set (https://github.com/gitbrent/PptxGenJS/blob/master/tsconfig.json#L4). 

I think that is supposed to output a types definition file when ts compiles which would mean the `types/index.d.ts` would not be needed. Will look into why that file is not being created in the `out` directory any longer.

cronin4392: @gitbrent I haven't had a chance yet to track down the `delcarationDir` issues. I think I could look into generating the types once in another PR.

gitbrent:
@cronin4392 - thanks for the patch and the research.

AFAIK, only an `index.d.ts` file is honored typescript for definitions, so i combine all the *.ts files into a single file and ship that.

src/core-interfaces.ts
```diff
@@ -302,7 +302,6 @@ export interface ITableOptions extends PositionOptions {
 	colspan?: number
 	colW?: number | number[]
 	fill?: Color
-	fontSize?: number
```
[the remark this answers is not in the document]
gitbrent: @cronin4392  - Perfect, this is much needed

gitbrent:
Thanks @cronin4392 !

PR #661, opened 2020-01-26
Update README.md

knowledgebaum:
Fixed Quickstart by adding require


gitbrent:
Thanks @knowledgebaum 

I've split the quick start guide into 2 sections and added a new one for ES6-style

Issue #678, opened 2020-02-21
UnhandledPromiseRejectionWarning (node) / Uncaught (in promise) (browser) when passing text as undefined 


gytisgreitai:
node code:
```
  try {
    slide.addText([{text: undefined, options: {breakLine: true}}], {x:1, y:1, w: 1, h:1})
  } catch (e) {
    console.log('this will not be invoked')
    console.error(e)
  }
```

Fails with:
```
(node:63649) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().
```
Example to reproduce in browser:

Fiddle: https://jsfiddle.net/6ayzoxh2/

Clearly this is an error on my side, and I should not pass the undefined text, but it would be cool if your code would throw, then I could handle it

ankon:
We're getting hit by this as well.

The problem indeed is that one cannot even catch this exception at all, so all our "defensive coding" didn't help.

I'm right now looking into this, and a very initial hunch is that the whole `save` code is failing to properly _reject_ the involved promises.

ankon:
Note that the error only happens when actually writing the output, not at the time you call `addSlide`. 

I agree that it would be useful to know _which_ slide had the issue, but that seems like a lot of error checking that would need to be added. 

The TypeScript definitions define that `text` must not be null/undefined, so if you use TypeScript *and* have proper types everywhere in your app there's a good chance that you'll be able to catch the error at development time rather than at runtime.

gitbrent:
This is fixed thanks to Pull #695

PR #695, opened 2020-03-06
Handle errors with promise rejections

ankon:
The code used promises, but in most cases dropped errors and rejects "on the floor", rather than just
exposing them further to the caller. This made it impossible to actually catch these problems, leading
to NodeJS warnings about "unhandled promise rejections" (which will soon make the NodeJS process
crash!)

---
This is related to https://github.com/gitbrent/PptxGenJS/issues/678, and at least allows to catch the errors.

The fixes follow a couple of patterns:

1. Remove useless wrapping promises when there is already one available

   ```js
   return new Promise((resolve, _reject) => doSomePromiseCode.then((result) => resolve(result)))
   ```

   This can be written as simply `return doSomePromiseCode`, which also makes the caller able to handle the rejection.

2. Move the use of `new Promise()` calls for callbacks as close to the callback as possible.

3. Make sure to return promises (for example the result of `Promise.all`)

Note that this does still contain the `Presenation` typo in `writeFile`, see #694 for that.

### DIFF
--- a/src/pptxgen.ts
+++ b/src/pptxgen.ts
@@ -395,150 +395,137 @@ export default class PptxGenJS implements IPresentation {
 	 * @param {Blob} blobContent - Blob content
 	 * @return {Promise<string>} Promise with file name
 	 */
-	private writeFileToBrowser = (exportName: string, blobContent: Blob): Promise<string> =>
-		new Promise((resolve, _reject) => {
-			// STEP 1: Create element
-			let eleLink = document.createElement('a')
-			eleLink.setAttribute('style', 'display:none;')
-			document.body.appendChild(eleLink)
+	private writeFileToBrowser = (exportName: string, blobContent: Blob): Promise<string> => {
+		// STEP 1: Create element
+		let eleLink = document.createElement('a')
+		eleLink.setAttribute('style', 'display:none;')
+		document.body.appendChild(eleLink)
 
-			// STEP 2: Download file to browser
-			// DESIGN: Use `createObjectURL()` (or MS-specific func for IE11) to D/L files in client browsers (FYI: synchronously executed)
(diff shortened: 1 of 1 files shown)


ankon:
I've tested this by updating our application to use a pptxgenjs distribution compiled from this PR, and seeing that we now correctly catch the errors caused by undefined `text` given to `addSlide` in the saving path (see #678 for that problem)

src/pptxgen.ts
```diff
@@ -560,15 +539,10 @@ export default class PptxGenJS implements IPresentationLib {
...
-				.then(content => {
-					resolve(content)
-				})
-				.catch(ex => {
-					reject(ex + '\nDid you mean to use writeFile() instead?')
-				})
-		})
+		return this.exportPresentation(outputType)
+			.catch(ex => {
+				throw new Error(ex.message + '\nDid you mean to use writeFile() instead?')
```
ankon: FWIW: This kills the stack trace of the actual exception, which seriously sucks for debugging. I would suggest to actually drop this as well, and just do:

```js
write(outputType: JSZIP_OUTPUT_TYPE): Promise<string | ArrayBuffer | Blob | Buffer | Uint8Array> {
	return this.exportPresentation(outputType)
}
```

ankon:
@gitbrent do you maybe have some time to look over this? Right now the PptxGenJS library kills our processes due to the unhandled rejections, and this patch seems to fix things nicely -- but of course we don't want to run with local patches if possible :)

ankon:
Anything I can do to help move this forward?

ankon:
Link to #719 (fixed also by #720).

ankon:
Rebased onto current master to resolve the conflict introduced through #694

gitbrent:
Thanks @ankon - great work! Appreciate the Promise expertise.

PR #700, opened 2020-03-18
Update README.md

lucidlemon:
minor typo regarding alignment of an object

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ let slide = pres.addSlide();
 
 // 3. Add one or more objects (Tables, Shapes, Images, Text and Media) to the Slide
 let textboxText = "Hello World from PptxGenJS!";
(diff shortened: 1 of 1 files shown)


ankon:
`pres` doesn't seem to be declared for the `ITextOpts` interface, and the constant name `AlignH.center` also seems to point to this being the `align` option.

https://github.com/gitbrent/PptxGenJS/blob/518667b3b23311275d9869846c90e1254a758c5f/types/index.d.ts#L1323-L1381

gitbrent:
This is not correct. `align` is the proper way to align text...

lucidlemon:
ah jesus my bad! i didn't mean to change the key - but the var `pptx` is undefined as you are referencing `pres`, fixed that now

gitbrent:
@lucidlemon - thanks for the fix!

Issue #705, opened 2020-04-09
catAxisTime options are not recognized


lucasflomuller:
I receive this warning on my console:
![image](https://user-images.githubusercontent.com/32915144/78935015-56463400-7a82-11ea-83e4-e44b301009b0.png)

But my code is:
![image](https://user-images.githubusercontent.com/32915144/78935069-6e1db800-7a82-11ea-8acc-eb270b29cf1c.png)

Apparently the code on that part looks okay, but I investigated a lot and I don't know what's going on. I really needed these properties to be functional to get my expected result.

Any ideas of what it is?

gitbrent:
@lucasflomuller - thanks so much for identifying and fixing this.

I've asked for a change on the pull request, otherwise, it looks good.

PR #706, opened 2020-04-09
fix time units validation (issue #705)

lucasflomuller:
fix issue #705.

It was a simple problem, this is my first contribution so if I didn't follow some standard I'm sorry.


gitbrent:
Thanks @lucasflomuller !

PR #710, opened 2020-04-13
Add support for current date as an element

sgenoud:
You can now define in master a `current-date` element!


lambrent:
will it be included in the master branch?

PR #720, opened 2020-04-27
Fix bluebird warning when executing pptx.stream()

### LINKED ISSUE #719: Bluebird warning when executing pptx.stream()
When pptx.stream() is executed, bluebird logs a warning:

> (node:10443) Warning: a promise was created in a handler at domain.js:137:15 but was not returned from it, see http://goo.gl/rRqMUw

This is because zip.generateAsync promise is not returned in this line:
https://github.com/gitbrent/PptxGenJS/blob/7654f1594a3204d913571fd75650e1990c6e75e4/src/pptxgen.ts#L526

When I add the return statement, the warning goes away.

tkounenis:
Hi there, fixing the issue reporting earlier today. Cheers!

Fixes #719

### DIFF
--- a/src/pptxgen.ts
+++ b/src/pptxgen.ts
@@ -523,7 +523,7 @@ export default class PptxGenJS implements IPresentation {
 						.then(() => {
 							if (outputType === 'STREAM') {
 								// A: stream file
-								zip.generateAsync({ type: 'nodebuffer' }).then(content => {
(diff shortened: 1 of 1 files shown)


ankon:
There's _a lot of these issues_: See https://github.com/gitbrent/PptxGenJS/pull/695

gitbrent:
Thanks @tkounenis - this has already been fixed thanks to Pull #695

Issue #663, opened 2020-01-28
Master slides 'bkgd' value overwritten by subsequent master slide definitions


readeral:
Using the latest master branch (not latest release on NPM), I'm unable to have two master slides with different images.

The following code results in the second declared 'bkgd' being the background for both types of slide:

```
  pptx.defineSlideMaster({
    title: 'TITLE_SLIDE',
    bkgd: {path: titlebgimg}
  });

  pptx.defineSlideMaster({
    title: 'LYRIC_SLIDE',
    bkgd: {path: mainbgimg}
  });
```

Not that I think it's relevant, but just in case, I am importing my background images using webpack loader file-loader and passing them as props to the component which is producing the pptx.

Is there a different way to ensure distinct bkgd values, or is this a bug?

nzerr:
FYI, this issue also occurs with the latest pptxgenjs NPM package (3.1.1). Is there any possible workaround yet?

dgavey:
I can verify that I am also seeing this issue on the latest (3.1.1).  I've confirmed that is an issue for both data blobs and for image URL's.

dgavey:
I was able to dig into this a bit more today.  It must be some sort of async issue.  If one of the background images is loaded locally and the other from a internet URL, it does appear to work, but not if I'm using two local paths or even two data blobs. 

I pulled down this project to try to find a fix, haven't found it yet though.

jrohland:
Found the underlying issue. When the master slide's image is added to the bundle it's being saved with the hard coded image filename of '../media/image-1.ext'. I created a PR #726 which includes the master slide's title as part of the image filename.

gitbrent:
Thanks to @jrohland for the fix.

PR #726, opened 2020-05-01
Add the slide layout name to the generated background image name

jrohland:
Currently when adding multiple master slides with background images, their background image is being overwritten with the last master slide's background image. Fixes issue #663


gitbrent:
Thanks @jrohland !

Issue #725, opened 2020-05-01
addShape(...rectangle,{..}) gives an invalid shape error...


Offbeatmammal:
If I code
```slide.addShape(pptx.ShapeType.hexagon, {x:0.5, y:0.8, w:1.5, h:3.0, fill:'FF0000'});```
then everything works as expected
But, if I use 
```slide.addShape(pptx.ShapeType.rectangle, {x:0.5, y:0.8, w:1.5, h:3.0, fill:'FF0000'});```
I get:
        ```pptxgen.min.js:2 Uncaught Error: Missing/Invalid shape parameter! Example: 
        'addShape(pptx.shapes.LINE, {x:1, y:1, w:1, h:1});'
         at Fe (pptxgen.min.js:2)
        at Oe.addShape (pptxgen.min.js:2)
        at ppt.html:19```

using ```<script src="https://cdn.jsdelivr.net/gh/gitbrent/pptxgenjs@latest/dist/pptxgen.min.js"></script>``` to get the current library

Offbeatmammal:
looking in source, ```rect``` is an acceptable shape name, but ```rectangle``` (from the examples) isn't (not sure if you want to address as a doc fix or add rectangle as a synonym?)

gitbrent:
Thanks @Offbeatmammal 

Documentation was recently updated to use `rect`

PR #741, opened 2020-05-18
Fixing shape example in documentation

thamara:
This actually fixes one of the issues brought up in #725.


gitbrent:
Thanks @thamara !

PR #746, opened 2020-05-19
#745 fix '0' printing in slides

### LINKED ISSUE #745: Add 0 as number to slide via "addText()" method.
Hi.

> https://github.com/gitbrent/PptxGenJS/blob/3dd977b200113559abb1ef80e899ac13c9d26858/src/gen-objects.ts#L828

I use method 'addSlide()' from class Slide to print numbers on presentation and found issue with '0'
If you put 0 as type number, it wouldn't print into slide.
Cause it handle '0' as false in if condition and replace it to empty string - ''

roman1bondarenko:
fix #745 issue

### DIFF
--- a/src/gen-objects.ts
+++ b/src/gen-objects.ts
@@ -815,17 +815,17 @@ export function addTableDefinition(
 
 /**
  * Adds a text object to a slide definition.
- * @param {string|IText[]} text
+ * @param {string|number|IText[]} text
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks for the contribution @roman1bondarenko 

However, i'm leaving `text` as a string type only as that's been the design for a while and matches `TableCell` text property type as well.

PR #761, opened 2020-06-05
Removed Buffer

### LINKED ISSUE #654: Integration with Angular 8
I'm fairly new to Angular so maybe I'm setting this up wrong, but I can't get pptxgenjs version 3.1.0 to work with Angular 8.3.0.

I made a plain Angular app using `ng new app` and then called `npm install pptxgenjs` to add it. When I run `ng serve` to build, I get a few errors like this:
``ERROR in node_modules/pptxgenjs/types/index.d.ts:85:50 - error TS2591: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.  ``

So then I add `node` to the types field in `tsconfig.app.json`. Any way I try to build from there, or change the import/require of pptxgenjs in my angular component, I always get this build error:
``ERROR in ./node_modules/pptxgenjs/dist/pptxgen.es.js
Module not found: Error: Can't resolve 'fs' in 'C:\projects\slides-generator\node_modules\pptxgenjs\dist'``

I looked at some other issues that have been closed, and maybe this is due to pptxgenjs thinking it's running on node instead of angular, but I don't know where that code is. (And I tried to put breakpoints in pptxgen.es.js everywhere I saw `fs` but they never got hit. Probably since the VS Code  debugger was looking at runtime code instead of the compile process).

Any thoughts on how I can work around this or fix this?

lustigerlurch551:
Removed `Buffer` to work with Angular
`Buffer` is a node.js datatype. It's not supported in Angular.

Without this fix you get the following errormessage, if you use it with angular:
`ERROR: node_modules/pptxgenjs/types/index.d.ts(67,50): error TS2591: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.`

fixes #654

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -64,15 +64,15 @@ declare class PptxGenJS {
 
 	/**
 	 * Export the current Presentation to stream
-	 * @returns {Promise<string | ArrayBuffer | Blob | Buffer | Uint8Array>} file stream
+	 * @returns {Promise<string | ArrayBuffer | Blob | Uint8Array>} file stream
 	 */
-	stream(): Promise<string | ArrayBuffer | Blob | Buffer | Uint8Array>
+	stream(): Promise<string | ArrayBuffer | Blob | Uint8Array>
 	/**
 	 * Export the current Presentation as JSZip content with the selected type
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks @lustigerlurch551 !

Issue #788, opened 2020-06-30
dataLabelPosition "t" in Bar chart is crashing ppt in latest MS office Power Point


jsvishal:
dataLabelPosition with option "t" crashing MS office PPT.

Chart Type: BAR

### Category
- [ ] Enhancement
- [ ] Bug 

### Version
3.3.0
version(s) of PowerPoint we are targeting: [   latest MS office      ]

### Expected / Desired Behavior / Question
dataLabelPosition with option "t" crashing MS office PPT.

### Steps to Reproduce
dataLabelPosition with option "t" 
Chart Type: BAR

ReimaFrgos:
`dataLabelPosition: "t"` is not supported by bar charts. "t" is for use with bubble charts.

Guessing what you are after is Outside End?
`dataLabelPosition: "outEnd"`

gitbrent:
Fixed via Pull #938

Issue #768, opened 2020-06-12
Setting dataLabelPosition to a line chart causes latest office application to ask for repair


artdomg:
### Category
- [ ] Enhancement
- [X] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [3.2.1 ]

Please specify what version(s) of PowerPoint you are targeting: [16.37 (2020)]

### Expected / Desired Behavior / Question
*PPTX files should be able to open on Microsoft PowerPoint*

### Observed Behavior
When adding a line chart with dataLabelPosition set, Microsoft Office application asks to repair the file and fails to do it.

### Code sample
`
slide.addChart(
      ppt.ChartType.line,
      [
        {
          name: 'Foo',
          labels: ['bar'],
          values: [1]
        }
      ],
      {
        dataLabelPosition: 'bestFit'
      }
    );
`

gitbrent:
There are lots of demo charts that utilize the `dataLabelPosition` option, so it's likely that 'bestFit' doesn't work well with Line Charts.

ReimaFrgos:
bestFit isn't a valid label position for line charts:

![image](https://user-images.githubusercontent.com/14120424/84733242-4857e680-afd0-11ea-9d84-16a6c6a3dd42.png)

It's used for Pie and Doughnut charts

gitbrent:
Fixed via Pull #938

PR #789, opened 2020-07-03
Chart Validation

ReimaFrgos:
@gitbrent The attached edits should help eliminate repair errors resulting from improperly configured charts. Configurable values are checked against valid options by chart type.

Should resolve errors referenced in:
#788 #768 

Should resolve previously worked around issues for Office Online.

Updates run successfully against all Chart demo's. The only slide change is the Bubble Chart demo on Slide 15, bottom left. The data labels shift from center to right as per default MS Office behaviour. This could be addresses by either tweaking the edit to default this to Center, or update the demo too set the position to center.

### DIFF
--- a/dist/pptxgen.es.js
+++ b/dist/pptxgen.es.js
@@ -3095,23 +3095,56 @@ function addChartDefinition(target, type, data, opt) {
     // B: Options: misc
     if (['bar', 'col'].indexOf(options.barDir || '') < 0)
         options.barDir = 'col';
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks @ReimaFrgos - would you mind patching the ./src file instead :)

ReimaFrgos:
@gitbrent closing this PR and replacing with #790 instead which updates the ./src files instead.

PR #794, opened 2020-07-06
Issues 788+768 - Validate dataLabelPosition is valid for charts

ReimaFrgos:
@gitbrent this PR replaces #789 and updates the ./src/*.ts files instead.
(This is a resubmit because of a config issue)

Edits will override user actions with default settings if they try to enter an invalid input for either a dataLabelPosition or barGrouping value. barGrouping affects valid data labels which is why this is included.

This should resolve errors raised in:
#788 #768

It should also resolve issues relating to Office Online usage.

I have removed redundant validation that is no longer required when inserting the dataLabelPosition string

Have built the library through npm scripts > ships and outputted .js file runs successfully against all demo scripts. There is one minor change to bubble chart behaviour as seen in the Bubble Chart demo on Slide 15, bottom left. The data labels shift from center to right as per default MS Office behaviour. This could be addresses by either tweaking the edit to default this to Center, or update the demo too set the position to center depending on how sever you consider this impact.

### DIFF
--- a/src/gen-charts.ts
+++ b/src/gen-charts.ts
@@ -751,10 +751,7 @@ function makeChartType(chartType: CHART_NAME, data: OptsChartData[], opts: IChar
 					strXml += '        </a:defRPr>'
 					strXml += '      </a:pPr></a:p>'
 					strXml += '    </c:txPr>'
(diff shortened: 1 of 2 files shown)


gitbrent:
Closing: Superseded by Pull #938

PR #807, opened 2020-07-24
Add void return type to addSection definition

cronin4392:
With latest code from `master` I was getting a type error that a return type was not explicitly set and was defaulting to `any`


eugene1g:
Looks like this has been implemented & merged in https://github.com/gitbrent/PptxGenJS/pull/840

gitbrent:
Thanks @cronin4392 - this has been fixed via another patch and will be in v3.4.0

PR #808, opened 2020-07-24
Fix data label position for Pie chart

cronin4392:
I ran into this code just by chance. Not sure how it did not error out before? I think this is what it was intended to do? Before the string was not being added to the `strXml`


gitbrent (approved):

gitbrent:
Thanks @cronin4392 !

Issue #285, opened 2018-02-10
Bar chart with one series displays broken if chartColors argument is given


andrei-cs:
Creating a bar chart with one series and providing a chartColors array seems results in a graph looking like this:

<img width="747" alt="screen shot 2018-02-10 at 20 36 16" src="https://user-images.githubusercontent.com/3409418/36065375-79590d26-0ea2-11e8-8819-19f281b099ae.png">

Removing the chartColors param displays it correctly but I can't find another way of setting the color.

The problem is exactly as in this issue - https://github.com/gitbrent/PptxGenJS/issues/225

Providing a single element in the chartColors array  solves the color problem but still creates a legend with all the values in the labels array.

Looking through the code it seems like it is somewhat of an intended behaviour:

https://github.com/gitbrent/PptxGenJS/blob/e4c46021f91deb4b34be9043e72d5f5ae37ab391/dist/pptxgen.js#L2691-L2695

How would one set a custom color for a bar chart ?

gitbrent:
Hi @andrei-cs,

I'm not quite sure what the chart you're trying to produce would look like, but `chartColors` is the way to pass colors.

Using an array of colors with a single array is a specialized feature, any "normal" chart would just use something like this:
```javascript
chartColors: ['0088CC', '99FFCC'],
```

![screen shot 2018-02-11 at 18 42 19](https://user-images.githubusercontent.com/7218970/36080777-55666d8c-0f5b-11e8-92ea-f80e1b46580f.png)

Have you looked through the chart code in `examples/pptxgenjs-demo.html` ?

andrei-cs:
Maybe I wasn't clear enough explaining my problem.

The following code: 

```
var slide = pptx.addNewSlide();
	var opts = { x:1.0, y:1.0, font_size:42, color:'00FF00' };
 	var data = [ { name: 'Sample',
    labels: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23 ],
    values: [ 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 2, 1, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0 ] } 
];

  slide.addChart(pptx.charts.BAR, data, {showLegend:true});
	pptx.save(); 
```

generates the graph below with the correct legend being the Series name

[...]

PR #821, opened 2020-08-07
Single data set with a custom color should not create legends for each category

tvt:
When creating a bar chart with a single data set and setting custom colors each category gets a legend-entry. See #285 for last time that happened.

I guess the idea is that if you pass in an array of colors to a single data set, that means that you want each category to have a different color.

I think it would make more sense if creating legend entries for categories were an option that could be set.
This solution instead add a check to avoid trying to color each category, if only a single color is passed.


tvt:
Hi @gitbrent, anything else you need from me to help get this merged?

gitbrent:
@tvt - got any test cases?

tvt:
@gitbrent sure:

```
const slide = pptx.addNewSlide();
var data = [
    {
        name: 'Sample',
        labels: [0, 1],
        values: [2, 5],
    },
];

slide.addChart(pptx.charts.BAR, data, { showLegend: true, chartColors: ['0000FF'] });
pptx.writeFile();
```

Gives the output in Powerpoint 2016:
screenshot_2020-12-28-105836

gitbrent (approved):

gitbrent:
Thanks @tvt !

PR #831, opened 2020-08-25
Added new axis line options to chart settings

twatson83:
Added support for setting axis line style, color and size.


gitbrent:
Thanks @twatson83 !

PR #832, opened 2020-08-25
Fix secondary axis param check

twatson83:
If a secondary axis is not defined in the settings it is possible for multiple charts to use the same val axis.  

This change updates the secondary param check to only throw an exception if a secondary axis is defined.


gitbrent:
Thanks @twatson83 !

Issue #827, opened 2020-08-21
[BUG] lineSpacing option set to decimal triggers repair alert


ReimaFrgos:
Thank you for reporting an issue, suggesting an enhancement, or asking a question. 

We appreciate your feedback - to help the team understand your needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [Docs](https://gitbrent.github.io/PptxGenJS/docs/installation.html) to see if your question is already addressed there. This will help us ensure our documentation covers the most frequent questions.**

### Category
- [ ] Enhancement
- [X ] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [ 3.3.0  ]

Please specify what version(s) of PowerPoint you are targeting: [ N/A ]

*If you are not using the latest release, please update and see if the issue is resolved before submitting an issue.*

### Expected / Desired Behavior / Question
Should be able to enter a decimal for lineSpacing, not just an integer.
If using fontSize 12 and you want 1.15 line spacing you need linespacing set to 13.8

### Observed Behavior
Entering decimal for the lineSpacing option triggers a repair warning. An integer works fine

### Steps to Reproduce
enter lineSpacing: 13.8 as an option

### Fix
Update
`strXmlLnSpc = "<a:lnSpc><a:spcPts val=\"" + textObj.options.lineSpacing + "00\"/></a:lnSpc>";
`
to 
`strXmlLnSpc = "<a:lnSpc><a:spcPts val=\"" + (textObj.options.lineSpacing * 100) + "\"/></a:lnSpc>";
`

gitbrent:
Thanks @ReimaFrgos !

PR #840, opened 2020-09-02
Fixes for several rounding mistakes to achieve better precision, accuracy, and usability

michaelcbrook:
First of all, really appreciate all your hard work on this library.

In the spirit of improving it and solving potentially many issues users could have with it (myself included), I've identified and fixed many instances of incorrect multiplication or rounding, which will resolve many common mistakes when passing decimal numbers into the API, and will actually increase the accuracy of the resulting PPTX to be perfectly consistent with the input style parameters.

The fixes:

1. **Font sizes**. Virtually all font sizes are using the following calculation: `Math.round(fontSize) + "00"`. While this doesn't cause the resulting file to be corrupt, it removes a great deal of precision. If the font size input is 12.5, the output becomes 13. Alternatively, all of these should be `Math.round(fontSize * 100)`. This will produce exactly the right result of 12.5.

2. **lineSpacing**. The above fix was applied partially to lineSpacing [recently](https://github.com/gitbrent/PptxGenJS/issues/827), however, lineSpacing was only multiplied by 100 but the result was not rounded. As a result, round-off errors could still occur. (i.e. 70.79 * 100 = 7079.000000000001 in JavaScript). Rounding additionally after the multiplication prevents this. This problem [bit me](https://medium.com/@michaelcbrook/e-commerce-sites-should-watch-out-for-this-bug-cd5b70bb87bb) a number of years ago. Unfortunately, if this fix isn't applied, there is no workaround for making this work.

3. **Rotations**. Similar to the above, the result was not being rounded, which made it susceptible to the round-off error. It also would fail if the input rotation was much more than four decimal places long.

4. **Transparencies**. Specifically, shapeFill alpha, shapeFill transparency, and chartColorsOpacity all were similar to # 1, simply being suffixed with "000" instead of properly being multiplied by 1000 and then rounded. Using decimals in these values would have simply not worked before, when they could completely be supported and produce a result with higher precision and accuracy.

Hopefully these changes will help prevent a lot of the hangups people first have when using it (not knowing which numbers accept decimals, which don't, how many decimals, etc.) and give everyone the ability to achieve near-perfect precision that was needlessly prevented before just because of the subpar multiplication logic.

Thanks again for the great library.


michaelcbrook:
Just wanted to make a note on this additional commit. It contains:

1) Additional places where rounding was missed but should be performed (that I found later).

2) An unrelated bug fix from lines [2163 - 2166](https://github.com/gitbrent/PptxGenJS/blob/67f5dbe29429b83f45f35dbc039644cb26352041/dist/pptxgen.cjs.js#L2163) that fixes a bug where, if a parent textbox is set to bold, italics, underline, or strike, words inside of the textbox can't be un-bolded, un-italicized, un-underlined, or un-striked by explicitly setting the word-level formatting to `false`. This is because the code only explicitly adds the associated attribute for the style if the value is truthy, but in cases where it's explicitly false, the attribute should be added with a negatory value in order to not just inherit the style from the parent textbox.

Thank you!

michaelcbrook:
Fixed! :+1:

gitbrent (approved):
Perfect

gitbrent:
Thanks @michaelcbrook - great work!

PR #857, opened 2020-10-21
text run background color (i.e. highlight)

wangfengming:
Mac has this option. Windows do not has this option. But highlight works for windows.

image

image


gitbrent:
Thanks @wangfengming !

Screen Shot 2021-04-26 at 22 53 25

Screen Shot 2021-04-26 at 22 49 14

PR #861, opened 2020-11-05
[FEAT] angleRange & arcThicknessRatio properties for shapes

apresmoi:
Hello everybody.

I wanted to have this two options to be able to play around with the arc, blockArc and pie shapes. In the current version there isn't the possibility to set the angle range of these shapes.

Maybe a better solution for the future could be adding a type for each of the supported properties with all of their supported variables.

Hope you like it.
Thank you !


apresmoi:
Hi @gitbrent, do you need me to fix those conflicts ?

gitbrent:
@apresmoi - yes, please patch the src/*. files instead of dist. thanks.

apresmoi:
@gitbrent - There it is, I've force-pushed only with changes inside src/

gitbrent (approved):

gitbrent:
Thanks @apresmoi !

PR #871, opened 2020-12-15
Decimal fontsizes and Breakline Fix

Martin-N:


gitbrent:
Needs some work: We cannot just remove `Math.round()`, we'll end up with a regression.

Martin-N:
I didn't just remove the `Math.round()`, I multiplied the font size by a hundred. The problem was that '00' was added at the end of the font size value. So if a decimal value came through it became 7.500 instead of 7500. By multiplying the font size by 100 fixes that issue. You can test it. It works fine.

Martin-N:
Sorry, new to git. Wanted to create a new pull request for the breakline fix, but I see it ended up in the decimal font fix.

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #886, opened 2021-01-15
Improve types

mmarkelov:
Just removed some redundant types and fix `fill` typing, cause it is can also be a string, not only an object


gitbrent (approved):

gitbrent:
Thanks @mmarkelov !

PR #954, opened 2021-05-20
Allow specifying a modified theme color

### LINKED ISSUE #953: Adding color modifier support for scheme colors
Thank you for reporting an issue, suggesting an enhancement, or asking a question. 

We appreciate your feedback - to help the team understand your needs please complete the below template to ensure we have the details to help. Thanks!

**Please check out the [Docs](https://gitbrent.github.io/PptxGenJS/docs/installation.html) to see if your question is already addressed there. This will help us ensure our documentation covers the most frequent questions.**

### Category
- [X] Enhancement
- [ ] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: 3.6

### Discussion

I'd like to add support for setting color modifiers such as lumMod, lumOff, satMod, shade, tint, and so on when referencing a scheme color. I'm planning to implement this very soon for my own use, so opening this issue for discussion as to how to structure the API when I create the PR.

The actual use case here is that I'm generating templates from existing powerpoint files and then using pptxgenjs to create a new Powerpoint, so it's useful to be able to reuse the same modifiers created by Powerpoint in the source presentations, rather than trying to replicate all those modifiers myself through color calculations.

All of these modifiers are child elements on the `schemeClr` element, as described at https://c-rex.net/projects/samples/ooxml/e1/Part4/OOXML_P4_DOCX_schemeClr_topic_ID0EBRNJB.html. 

My current plan is to do something like this in index.d.ts:

```typescript
export type HexColor = string
export type ThemeColor = 'tx1' | 'tx2' | 'bg1' | 'bg2' | 'accent1' | 'accent2' | 'accent3' | 'accent4' | 'accent5' | 'accent6'
export interface ModifiedThemeColor {
  color: ThemeColor;
  shade?: number;
  tint?: number;
  satMod?: number;
  gray?: boolean;
  // And all the rest...
}
  
export type Color = HexColor | ThemeColor | ThemeColorWithModifiers
```

The percentage modifiers will follow the existing convention used by the `transparency` argument of a number from 0 to 100, which is then transformed to the internal "string of thousandths" format when writing the XML. I think this change will only really require updating the typescript definitions and the `createColorElement` function, so the impact *should* be fairly small, I think.

Any thoughts?

dimfeld:
Resolves #953.

This allows specifying all the modifiers that Powerpoint uses internally on theme colors, which is useful when using a powerpoint file as a template for output generated with pptxgenjs.

Haven't updated the docs yet. I wanted to file this draft PR to et any feedback on the design itself first.

### DIFF
--- a/src/core-interfaces.ts
+++ b/src/core-interfaces.ts
@@ -78,7 +78,39 @@ export interface BackgroundProps extends DataOrPathProps, ShapeFillProps {
  */
 export type HexColor = string
 export type ThemeColor = 'tx1' | 'tx2' | 'bg1' | 'bg2' | 'accent1' | 'accent2' | 'accent3' | 'accent4' | 'accent5' | 'accent6'
-export type Color = HexColor | ThemeColor
+export interface ModifiedThemeColor {
+	baseColor: HexColor | ThemeColor
+
+	alpha?: number
+	alphaMod?: number
+	alphaOff?: number
+	blue?: number
+	blueMod?: number
+	blueOff?: number
+	green?: number
+	greenMod?: number
+	greenOff?: number
+	red?: number
+	redMod?: number
+	redOff?: number
+	hue?: number
+	hueMod?: number
+	hueOff?: number
+	lum?: number
+	lumMod?: number
+	lumOff?: number
+	sat?: number
+	satMod?: number
+	satOff?: number
+	shade?: number
+	tint?: number
+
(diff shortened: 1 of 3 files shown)


matt88120:
Hi,

Looks like this is what i'm searching for. Do you have a exemple of use ?

PR #955, opened 2021-05-20
Use encodeXmlEntities for formatCode attributes

dimfeld:
The format codes generated by Powerpoint often contain quotes, and so this ensures that they generate valid XML.

e.g. the default currency formatter used by Powerpoint is `_(" $"* #,##0.00_);_(" $"* \(#,##0.00\);_(" $"* " -"??_);_(@_)`


gitbrent:
Thanks @dimfeld !

Issue #970, opened 2021-06-27
invertedColors: not working any more


ericlaaper:
HI, I have got a problem with the inverted colors in bar chart

Everything works except for this. In the past, same code worked perfect.
Underneath my code for the options off the graph

let optionsGraph = {
        x: 0.25,
        y: 0.85,
        w: 7,
        h: 4,
        barDir: "bar",
        barGrouping: "stacked",
        valGridLine: {
          color: "EFF0F1",
        },
       catAxisOrientation: "minMax",
        catAxisLabelColor: this.colors.colorB,
        catAxisTitleColor:this.colors.colorA,
        catAxisLabelFontFace: "Calibri",
        catAxisLabelFontSize: 14,
        catAxisLineShow: false,
        valAxisLabelFontColor: this.colors.colorA,
        valAxisLabelFontBold: false,
        catAxisLabelPos: 'low',
        dataLabelColor: this.colors.colorA,
        showValue: false,
        chartColors: [this.colors.colorC],
        invertedColors: [this.colors.colorD],
        titleColor: "red",
        titleFontFace: "Arial",
        titleFontSize: 24,
        valAxisMaxVal: 100,
        valAxisMinVal: -100, 

};
Then I add the chart by:

slide1.addChart(pptx.charts.BAR, arrayStatementA, optionsGraph);
the optiongraph is stating an error in my IDE  which removes when I remove InvertedColors form options.
removing the bracet from InvertedColors makes the error go away but it not makes InvertedColors working.

I also tried with just colornumber in stead off the variable same problem negative values have the same color as positive values.

ericlaaper:
![Schermafbeelding 2021-06-27 om 22 08 23](https://user-images.githubusercontent.com/20456800/123558009-510efb00-d794-11eb-8786-13c668feb399.png)


this is the chart what is produced

ericlaaper:
HI there, Any news on this?

ReimaFrgos:
I'm assuming you were using version 3.3.1 or older since you mentioned it used to work.

As far as I can tell this was broken in 3.4.0 as part of PR #821

Pretty sure the fix is to change the last line here to > 0 so that you can have a single element array.

```
 if ((chartType === CHART_TYPE.BAR || chartType === CHART_TYPE.BAR3D) &&
data.length === 1 &&
opts.chartColors !== BARCHART_COLORS &&
opts.chartColors.length > 1) {
```

As an interim fix, you can just create an array with the same color twice:
```   
[...]

PR #1004, opened 2021-09-30
Fix for `InvertedColors` (Issue #970)

leonyah:
invertedColors should be defined as "HexColor[]";
bar chart did not render inverted colored negative bars unless more than one color was provided in the chartColors array


gitbrent:
Thanks @leonyah !

Issue #339, opened 2018-05-18
Slide Number option in pptx.defineSlideMaster() not working


harshalgite:
Slide numbers are not being displayed for any slide in the Powerpoint.
I used the following code to define master slide:
`pptx.defineSlideMaster({
    title:'Title',
    slideNumber: {x: 0.3, y:7.15},
    objects: [...........]
});`

gitbrent:
Please ensure you:
* have the latest version 
* are defining and referencing the Master Slide correctly
* are not placing the slide number outside of the slide bounds

**Slide Number example:**
```
var pptx = new PptxGenJS();

pptx.defineSlideMaster({
    title: 'MASTER_SLIDE',
    slideNumber: { x:1.0, y:5.0 }
});

var slide = pptx.addNewSlide('MASTER_SLIDE');

slide.addText("Slide number test", { x:1.5, y:1.5, w:6, h:2 } );

pptx.save('PptxGenJS-Issue-339');
```

![screen shot 2018-05-29 at 23 06 42](https://user-images.githubusercontent.com/7218970/40698539-1e034e80-6395-11e8-9826-1781f4ce8bf3.png)

Your `y` value of `7` may be outside the viewable area:
![screen shot 2018-05-29 at 23 05 57](https://user-images.githubusercontent.com/7218970/40698593-613167b4-6395-11e8-979b-5ba6e506a129.png)

manokaran68:
Hi,

I tried the above code. I see a dashed box where the slide number is supposed to be but the slide number itself is not shown! Version am using is 2.3.0.20180912

NitinkumarPP:
Same issue occurring to me. Any update on this?

pingrosso:
Same issue as manokaran68. pptxgenjs 3.2.1

Aakarshak1:
Any update on this? I'm having same issue as @manokaran68. using version 3.1

leahjwoodhead:
I'm also having this issue! Is there an update yet, or another way around getting page numbers on the slides? Using  v 3.4.0

gitbrent:
There's no test case i'm aware of where page numbers do not show.

jessedp:
Using v3.6.0 -

In LibreOffice Impress, there is an empty, dashed box where the Slide Numbers are placed like @manokaran68  mentioned above.

In Powerpoint they are displayed.

The impact for me is that I actually want a pdf - so I'm running the pptx this generates through `unoconv` which uses LibreOffice.

PR #1016, opened 2021-10-20
Allow slide number to be set as bold

mathbruyen:
Handle bold from TextBaseProps by setting `b="1"` in the XML. Move up align definition before the field as it was no longer working when combined with bold.

Add actual slide number in the XML so that LibreOffice displays the slide number too (actually solves #339).


gitbrent:
Thanks @mathbruyen !

PR #1017, opened 2021-10-25
bugfix: each data marker on a doughnut as a different color

mathbruyen:
Doughnut charts display a single "Serie 1" label in their legend on
OpenOffice. This is due to `varyColor` (or "Vary Colors By Point") being
set to 0 on them. Since doughnut charts always have one color per
"point" (area), this should always be set to 1 for them.


gitbrent (approved):

gitbrent:
Thanks @mathbruyen !

PR #1021, opened 2021-10-29
allow chart legend to specify a manual layout

mathbruyen:
In some cases the automatic legend position from Powerpoint is not very good (it overlaps with the actual chart), thus allowing to specify a legend layout in the same way as chart layout can be specified.

### DIFF
--- a/src/core-interfaces.ts
+++ b/src/core-interfaces.ts
@@ -1295,6 +1295,7 @@ export interface IChartPropsLegend {
 	legendFontFace?: string
 	legendFontSize?: number
 	legendPos?: 'b' | 'l' | 'r' | 't' | 'tr'
+	legendLayout?: PositionProps
 }
 export interface IChartPropsTitle extends TextBaseProps {
 	title?: string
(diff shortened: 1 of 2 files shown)


gitbrent:
@mathbruyen - please provide some examples: source code and powerpoint output

mathbruyen:
We use this option to precisely control space used by both chart and legend. Using legend position alone was not using space as efficiently as we wanted, Powerpoint did not take full width and instead legend was overlapping with chart.

For example taking full width and respectively 66%/29% of the height with a 5% margin at the middle:

```js
const chartOptions = {
  layout: {
    x: 0,
    y: 0,
    h: 0.66,
    w: 1,
  },
  legendLayout: {
    x: 0,
    y: 0.66 + 0.05,
    h: 1 - 0.66 - 0.05,
    w: 1,
  },
};
```

image

mathbruyen:
Removed `<c:layoutTarget val="inner" />` that was causing Powerpoint to report invalid content on bar charts.

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1024, opened 2021-11-05
Fix some media bugs & support big video file 

canwdev:
- Fix: addMediaDefinition ext and cover
- Support big video files: [demo](./test/big_file.js)

```javascript
slide.addMedia({
	x: 1.0,
	y: 2.2,
	w: imgWidth,
	h: imgHeight,
	type: "video",
	path: videoUrl,
	ext: "mp4", // fix file extension name, Sometimes the url does not contain the suffix name
	cover: imgData.base64, // video cover base64 string
	isFsPath: true, // Let node.js read the file directly instead of converting to base64
});
```
```javascript
// Use stream output directly
const fileData = await pptx.stream({
	compression: false,
});
console.log("createWriteStream...");
const out = fs.createWriteStream(savePath);
fileData.pipe(out).on("finish", () => {
	console.log("Success!");
});
```


gitbrent:
Thanks @canwdev !

PR #1025, opened 2021-11-05
allow bullets to be cancelled by an arbitrary size

mathbruyen:


gitbrent:
@mathbruyen - can you provide some screencaps and description of what this accomplishes? I don't see where this does anything other than what the `indent` prop does...

mathbruyen:
Indent is specifying the position of the bullet relative to the beginning of the text.

There are two settings for list items. `marL` defines the distance between text block border and beginning of text. `indent` defines the position of the bullet relative to beginning of text. The current code is setting `marL = bulletMarL * indentLevel` to moves the text away from the border N times indentation, and `indent = -bulletMarL` which places the bullet at one level of indentation less than the text. By allowing to specify a different value for indent one can control the distance between bullet and beginning of text.

With current code the 1st bullet would appear below the "d", and the second below the "e", but with the ability to control it we slightly moved bullet to the right:

image

gitbrent:
Currently, users alter the left margin prop, to move the bullet itself over as this matches the available options in the powerpoint app. I still don't see what value this change brings.

mathbruyen:
Didn't realize the trick was possible. In our case it is not practical because we create text blocks with a mix of text and lists. The alternative would be to create multiple stacked text blocks, but it would hinder the ability to modify content manually afterwards (initial content is auto generated but we customize conclusions manually afterwards).

We don't actually use them but it would also give more control for nested lists. Currently bullets appears right below the beginning of text at previous indent (on the screenshot above bullets would be right below the "e" of "exemple"), with this change one can choose where it appears (on the screenshot I moved them slightly to the right).

gitbrent:
Closing as I dont see a larger audience for this.

The library currently supports indent levels and bullet spacing.

Screen Shot 2022-05-01 at 22 57 05

Issue #1008, opened 2021-10-06
PPT repair issue for long text.


Naveencheekoti17:
I am facing the issue with long text for tables. Text is having more than 150 characters

```javascript
var pptx = new PptxGenJS();
var slideIndex = pptx.addSlide();
var projRows = [
    [
        { text: "id", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", align: "left", valign: "middle" } },
        { text: "First item Desc", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Impact", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Owner", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Created Date", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Due Date", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Status", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
        { text: "Update", options: { fontFace: "Segoe UI", fontSize: 7, bold: true, fill: "1F3864", color: "ffffff", valign: "middle" } },
    ],
    [
        { text: "1", options: { fontSize: 7, align: "left", valign: "middle" } },
        {
[...]

PR #1028, opened 2021-11-11
Updated: Brand-new algorithm for table auto-paging

gitbrent:
Fixes Issue #1008 and makes addition of lines to each cell much more accurate.


gitbrent (commented):
Looks good

PR #1032, opened 2021-11-15
allow specifying bar overlap percentage in bar charts

mathbruyen:
By default bars in a bar chart are stacked together.

image

There is an option to set them on top of each other, but none to set how far from each other they should be. Powerpoint gives the ability to do so with "Series overlap":

image

With a value of -20% this gives the following:

image

Positive values can be used too (+20%):

image

The PR adds an option in bar chart options to control this value independently.


gitbrent:
Duplicate of Pull #1010

gitbrent:
Implemented via Pull #1010

PR #777, opened 2020-06-21
added new connector type

lifejuggler:
added new options to allow users to add connector between shapes

PR #1059, opened 2022-02-02
added connected lines for shapes

fred441a:
hi, i am aware of the [other pull request](https://github.com/gitbrent/PptxGenJS/pull/777) by @lifejuggler. But seeing as how that is over a year old i thought i would just contribute again. i have updated some of @lifejugglers code to be able to merge into the newest master branch, and added a bit of a patch solution to id's

```typescript
		if (slideItemObj.options != undefined) {
			if (slideItemObj.options.sId != undefined) {
				if (IDs.indexOf(slideItemObj.options.sId) > -1) {
					throw "ID is already in use, object / shape id cannot be the same for multiple objects / shapes";
				} else {
					IDs.push(slideItemObj.options.sId);
				}

			} else {
				if (IDs.indexOf(idx + 2) > -1) {
					throw "an sID used matched an automatically generated ID try using a higher number";
				} else {
					IDs.push(idx + 2);
				}

			}
		}
```
this at least throws an error if you have colliding id's, however i was wondering you you though it would be smarter to just automatically increment over already taken ids or if functions should return an automatically generated id to the user for them to use.

i also added an anchor enum that works at least for rectangular shapes. and added connectors to the shapes demo, like you asked.

i can also add connectors to other objects like tables or media, but i don't know if that is even somthing that we want?

should the user need to calculate their own position of of the connector arrows, or do we want this functionality built into the library? i can add this feature for squares at least.

i am very willing to help out with development of this feature if you just tell my what you think is best.

### DIFF
--- a/demos/modules/demo_shape.mjs
+++ b/demos/modules/demo_shape.mjs
@@ -96,6 +96,7 @@ function genSlide01(pptx) {
 	});
 
 	slide.addShape(pptx.shapes.RIGHT_TRIANGLE, {
+		sId: 20,
 		x: 0.4,
 		y: 4.3,
 		w: 6.0,
@@ -105,6 +106,7 @@ function genSlide01(pptx) {
 		shapeName: "First Right Triangle",
 	});
 	slide.addShape(pptx.shapes.RIGHT_TRIANGLE, {
+		sId: 21,
 		x: 7.0,
 		y: 4.3,
 		w: 6.0,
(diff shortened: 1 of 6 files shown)


Hoogkamer:
This is a great and very important feature to have! Will it be available soon?

gitbrent:
Thanks @fred441a - can you provide sample usage code?

fred441a:
i have added the connected lines for shapes functionallity to the shapes demo

f3ve:
Would love to see this feature added soon. Anything I can do to help with this process?

Michael-Reich:
Why isn't that implemented?

roberrrt-s:
Aware i'm necrobumping but I would really appreciate this functionality :)

leospalteholz:
Would love to have this feature. Currently generating an org chart in powerpoint and this would be a major upgrade with proper connectors instead of just unconnected lines.

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

klausjagd:
Thank you very much for the information. I really appreciate it. Will switch to the plus version soon :-)

PR #1100, opened 2022-04-13
serie name on bubble chart, category axis position, leader lines on bubble chart

mathbruyen:
## Ability to add serie name to bubble chart

The goal is to actually display a label aside each bubble. Adding a label per bubble requires generating a lot of XML: adding a new column with label as data, and data references to these labels.

Adding serie name is much easier, only setting a flag. The trick is then to create one serie per bubble: X axis contains the positions of all bubbles, then each serie contains a single value. It also allows setting the color for each bubble, again without complex generation.

| X Axis | Serie 1 | Size 1 | Serie 2 | Size 2 | Serie 3 | Size 3 |
|--------|---------|--------|---------|--------|---------|--------|
| 1.5    | 2.5     | 100    |         |        |         |        |
| 2      |         |        | 1.2     | 50     |         |        |
| 2.3    |         |        |         |        | 3.2     | 75     |

## Category axis position

Values axis can be set at an arbitrary position using `valAxisCrossesAt`, but category axis could not be positioned the same way. Add `catAxisCrossesAt` that behaves the same way to set it, and preserve previous behaviour when not set.

## Leader lines on bubble chart

Leader lines could be enabled by a property, but were not enabled on bubble chart. Those are not displayed by default as Powerpoint judges that when the label is at its default position, then leader line is redundant. But if the label is moved, leader line appears.

## Result

image

### DIFF
--- a/src/core-interfaces.ts
+++ b/src/core-interfaces.ts
@@ -1101,6 +1101,7 @@ export interface IChartPropsBase {
 	showLeaderLines?: boolean
 	showLegend?: boolean
 	showPercent?: boolean
+	showSerName?: boolean,
 	showTitle?: boolean
 	showValue?: boolean
 	/**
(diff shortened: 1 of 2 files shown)


gitbrent:
Thanks Mathieu!

PR #1108, opened 2022-05-11
Support bubble 3D chart type

mathbruyen:
Support bubble chart with a 3D appearance for bubbles instead of a solid display.

Implementation is done as a new chart type as Powerpoint UI defines it as a different type and bar 3D is defined as its own type as well. I'm not sure it is the best way, especially as even though Powerpoint UI treats it as a different chart type, in the XML it is a bubble chart with an option set to `1`. An alternative could be with a property (like `bubble3d: boolean` or `bubbleAppearence = 'solid' | '3d'`) but there is no existing type for bubble chart specific options.

If you prefer that setting to be done in any way feel free to tell and I will update the PR accordingly.


gitbrent:
Thanks @mathbruyen - great work!

Screen Shot 2022-05-15 at 12 39 51

PR #1122, opened 2022-06-07
bugfix: escape object name in chart xml

mathbruyen:
Object name is escaped at one place, but missing on another.


gitbrent (approved):

gitbrent:
Thanks @mathbruyen !

PR #1126, opened 2022-06-29
feature: allow selecting round or square line cap on line charts

mathbruyen:
Makes both the line and its legend having one of:

* flat line cap: ends with a straight line exactly at the end point
* rounded line cap: ends with a semi circle centred at the end point
* square line cap: ends with a straight line past end point by 1/2 of line width

Default is flat to preserve current behaviour.


gitbrent (approved):

gitbrent:
Thanks @mathbruyen !

PR #1140, opened 2022-08-01
fix #1139

### LINKED ISSUE #1139: autpPage duplicates text when text array is used
### Category
- Bug

### Version

Please specify what version of the library you are using: [ 3.10.0 ]

Please specify what version(s) of PowerPoint you are targeting: [Office 2010]

Please specify what web browser you are using: [node.js]

### Expected / Desired Behavior / Question
The text should be formatted correctly as it is passed in.

### Observed Behavior
The text is duplicated in a table cell when `autoPage: true` is used and a text array is supplied

### Steps to Reproduce
Create a table with the following input to observe the issue.
```javaScript
slide.addTable(
  [
    [
      {
        text: "column 1 header",
        options: { bold: true },
      },
      { text: "column 2 header", options: { bold: true } },
    ],
    [
      {
        text: [
          { text: "this will be duplicated" },
          { text: "1", options: { superscript: true } },
        ],
      },
      { text: "column 2" },
    ],
  ],
  { autoPage: true }
)
```

Or check out my example repo [here](https://github.com/mikemeerschaert/pptxgen-autopage-dupe-demo)

Here is an example screen shot to demonstrate what I'm seeing:
Screenshot 2022-08-01 115200

mikemeerschaert:
Fixes #1139.

This issue appears to be caused by the `parseTextToLines` routine when it is grouping teble cells to handle line breaks. There is a comment to flush the line buffer, but it just pushes more lines to the inputLines1 array without resetting the newLine buffer, resulting in the duplication when an array is supplied in the text field without a breakline parameter.


gitbrent:
Thanks @mikemeerschaert 

Patches need to be against the code src/*.ts files. Files under `dist` are generated. 

See other recent pull requests for examples.

mikemeerschaert:
Sorry about that @gitbrent, I removed the auto-generated files.

gitbrent:
@mikemeerschaert - can you provide a test case demonstrating the issue being patched?

mikemeerschaert:
@gitbrent I linked an example repo with a test case in #1139, you can see the bug in [this repo](https://github.com/mikemeerschaert/pptxgen-autopage-dupe-demo)

treystout:
bumping for interest. Hitting this exact case in latest. It's a bummer losing individual options per text node,  what does this request need for merge?

mikemeerschaert:
@treystout it looks like this is in the [3.12.0 milestone](https://github.com/gitbrent/PptxGenJS/milestone/39), so I think we just need to wait for @gitbrent to find the time to release that version.

gitbrent:
Thanks @mikemeerschaert !

PR #1143, opened 2022-08-05
Update image API doc

seekuehe:
The `transparency` property is available for images but is missing from the API doc. Info taken from `core-interfaces.ts`


gitbrent:
Thanks @seekuehe !

PR #1153, opened 2022-08-23
Update text docs: transparency + wrap

tjinauyeung:
Add missing properties to text API docs:
- `transparency`
- `wrap`

These are already supported by the lib, but were still missing from the documentation.


gitbrent (approved):

gitbrent:
Thanks @tjinauyeung !

Issue #1048, opened 2021-12-29
[BUG] PowerPoint shows "Repair" dialog when auto-paginated cell contains empty value


Barrywww:
### Category
- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Version

Please specify what version of the library you are using: [ 3.9.0 ]

Please specify what version(s) of PowerPoint you are targeting: [ PowerPoint for Mac v16.56 ]

### Expected / Desired Behavior
The powerpoint should open normally with no warning. 

### Observed Behavior
Powerpoint shows a "repair" dialog. 

### Possible Clues?
The xml of the problematic slide contains two different representation for empty cells. I suspect this may cause the problem. Here is a picture for reference:
![image](https://user-images.githubusercontent.com/36621849/147630995-8f553d7b-0ef1-42ec-94c8-e0a0b0f1b59d.png)

```XML
<a:txBody>
    <a:bodyPr />
    <a:lstStyle />
    <a:p>
        <a:endParaRPr lang="en-US" sz="800" dirty="0" />
    </a:p>
</a:txBody>
```
This pattern corresponds with the cell under Col1 in the overflowed line.

AND

```XML
<a:txBody>
    <a:bodyPr />
    <a:lstStyle />
</a:txBody>
```
This pattern corresponds with the cells under Col3 and Col4 in the overflow line.

The latter one **does not include <a:p> tag.** All of the slides contain the latter pattern cannot be properlly rendered without repairing.

The problem seems to be related to the length of the content overflowing. If there are only several words overflowing, than it is very likely to triggert the dialog. Please check the pictures below.

This one TRIGGERS the "Repair" dialog. And less words overflowing TRIGGERS.
![image](https://user-images.githubusercontent.com/36621849/147630995-8f553d7b-0ef1-42ec-94c8-e0a0b0f1b59d.png)

This one DOES NOT. And more words overflowing DOES NOT trigger.
![image](https://user-images.githubusercontent.com/36621849/147630806-a28afded-a991-4baf-95f5-f8d7c88a9020.png)

### Steps to Reproduce
Please feel free to change the length of the long text if the snippet below does not trigger the problem.

```JavaScript
const testExport = async () => {
  let pptx = new pptxgen()
  pptx.layout = 'LAYOUT_WIDE'
  let slide = pptx.addSlide()
  let tableRows = [
    [
      {text: "Col 1"},
      {text: "Col 2"},
[...]

PR #1160, opened 2022-09-05
fix: prevent empty txBody in autopaged empty cells

raulneis:
Whenever autopage generates a new slides, there might be empty cells, that need to be properly generated to avoid the 'repair' message in PowerPoint (here's the issue: https://github.com/gitbrent/PptxGenJS/issues/1048).

Here is the content generated before this fix for empty cells in rows broken into 2 or more pages
```xml
<a:tc>
  <a:txBody>
      <a:bodyPr/>
      <a:lstStyle/>
                              <----- missing empty p
  </a:txBody>
  ...
</a:tc>
```

This triggers the 'repair' message in PowerPoint.

Here is what PowerPoint generates after repairing the file:

```xml
<a:tc>
  <a:txBody>
    <a:bodyPr/>
    <a:lstStyle/>
    <a:p>
      <a:endParaRPr/>
    </a:p>
  </a:txBody>
  ...
</a:tc>
```

### DIFF
--- a/src/gen-xml.ts
+++ b/src/gen-xml.ts
@@ -1322,6 +1322,11 @@ export function genXmlTextBody (slideObj: ISlideObject | TableCell): string {
 		// D: End paragraph
 		strSlideXml += '</a:p>'
 	})
+	
+	// PowerPoint shows repair message if txBody misses p #1048
+	if (!arrLines.length) {
+		strSlideXml += '<a:p><a:endParaRPr /></a:p>'
+	}
 
 	// STEP 7: Close the textBody
 	strSlideXml += slideObj._type === SLIDE_OBJECT_TYPES.tablecell ? '</a:txBody>' : '</p:txBody>'


mdbaehre:
FYI - this also fixes the bug I reported:  Table Auto-Paging Error #1168

AdrianRS22:
@gitbrent any update on this one?

burivuhster:
I can confirm, this issue reproduces in one of my projects. Steps to reproduce are exactly what @mdbaehre described in #1168 

@gitbrent from my investigation, the root cause is getSlidesForTableRows function. There is a while loop, and under the described conditions the loop (Step 6 E) stops iterating on the last line of the cell containing text (https://github.com/gitbrent/PptxGenJS/blob/035cf6b26239e2ea1f27eefbcdb8134e504eb4f4/src/gen-tables.ts#L486) so the logic of adding empty text object for the remaining empty cells is not working.

My fix was very similar to the one proposed by @raulneis - just handle this on the XML building stage to ensure there will be no <a:txBody> without  <a:p>. I was going to submit a PR, but found this one.

As a temporary solution, we use a local fork of the library now, but I hope this can be merged into the main repo.

One more side note: I was able to find the issue with XML validation of slide files against XSD from https://ecma-international.org/publications-and-standards/standards/ecma-376/ (Part 4, Transitional schemas). I think it would be great to add unit and/or e2e tests for a variety of use cases, with validating library output against XSD schemas.

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1176, opened 2022-10-25
Prevent defineSlideMaster from mutating settings object that is provided by the user.

### LINKED ISSUE #1175: [BUG] Placeholder items end up in top left of slide in Lambda environment.
I just spent a couple of days tracking down a very hard to find bug, and now that I have a workaround, I wanted to write it up here so that:
1. Hopefully this can be addressed
2. Others can use this workaround and save themselves the headache.

Basically, the issue was that sometimes titles weren't being placed in the placeholders correctly when my function is being executed in the AWS Lambda environment, but it always worked when I was executing them locally, and it would sometimes work on AWS. I finally narrowed it down and realized the first execution after a cold start worked, but subsequent executions had the problem.

I spent considerable time print debugging through this library's source and re-deploying to track down the root cause, and after a few days of debugging I finally found out that `pptx.defineSlideMaster(MASTER_SETTINGS)` will mutate the `MASTER_SETTINGS` object that gets passed in by reference. It adds a few properties, but crucially it deletes the object's `options.name` and `options.type` for placeholders. This is probably fine most of the time, but if you are using a Lambda function and defining your `MASTER_SETTINGS` outside of the lambda function handler, the lambda environment will share that object between executions until the next cold start, meaning all subsequent executions will have the mutated object, and the library won't be able to apply the placeholders because `placeholder.options.name` has been deleted.

Here's the lines of code that are mutating the settings object. I don't know what the downstream effects of removing these lines are, so I'm just reporting the bug rather than submitting a PR for this one.
https://github.com/gitbrent/PptxGenJS/blob/a18a36a1a597e1130417741ea4d8d9363f7f5b48/src/gen-objects.ts#L79

Fortunately, there is a very simple fix for this once you know what is going on, just deep clone the settings object before using it so you get a fresh copy with every invocation.

```javaScript
const MASTER_SETTINGS_CLONE = JSON.parse(JSON.stringify(MASTER_SETTINGS))
//now use the cloned settings instead
pptx.defineSlideMaster(MASTER_SETTINGS_CLONE)
```

### Issue Category

- [ ] Enhancement
- [x] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Product Versions

- Please specify what version of the library you are using......: [3.11.0]
- Please specify what version(s) of PowerPoint you are targeting: [Office 2010]
- Please specify what web browser you are using.................: [node.js]

### Desired Behavior
The following code should place the title text inside the placeholder defined in MASTER_SLIDE
```javaScript
// src/libs/slideSettings/masterSlide.ts
export const MASTER_SETTINGS: pptxgen.SlideMasterProps = {
  title: "MASTER_SLIDE",
  background: { color: "FFFFFF" },
  objects: [
    {
      placeholder: {
        options: {
          name: "title",
          type: "title",
          x: 0.5,
          y: 0.6,
          h: 0.4,
          w: 10,
          margin: [0, 0, 7.0, 0],
          color: "003E36",
          fontFace: "+Heading",
          bold: true,
          fontSize: 18,
          valign: "top",
          align: "left",
        },
        text: "Title",
      },
    }
  ]
}
// src/libs/reporting/Report.ts
const pptx = new pptxgen()
pptx.defineSlideMaster(MASTER_SETTINGS)
const slideBuilder = pptx.addSlide({ masterName: "MASTER_SLIDE" })
slideBuilder.addText("I'm a title", {
  placeholder: "title",
})
```

### Observed Behavior
The title ends up in the top left of the slide at x:0, y:0 instead of where it's placed, but only in an AWS lambda function, and only on subsequent executions. The first execution, (from a 'cold start') works fine, but subsequent executions don't use the placeholders.
Also, the MASTER_SETTINGS object that gets passed in gets mutated, so a simple fix could be to simply deep clone the settings object before using it.

### Steps to Reproduce
1. Create a lambda handler that builds a presentation using a master slide with a placeholder
2. define your MASTER_SLIDE settings in a separate file
3. execute the function twice and save both presentations
The first execution should look correct, but the subsequent executions you'll see the text is in the top left of the slide at X:0, Y:0

mikemeerschaert:
This should fix #1175

### DIFF
--- a/src/pptxgen.ts
+++ b/src/pptxgen.ts
@@ -725,31 +725,33 @@ export default class PptxGenJS implements IPresentationProps {
 	 * @param {SlideMasterProps} props - layout properties
 	 */
 	defineSlideMaster (props: SlideMasterProps): void {
-		if (!props.title) throw new Error('defineSlideMaster() object argument requires a `title` value. (https://gitbrent.github.io/PptxGenJS/docs/masters.html)')
+		//deep clone the props object so we don't mutate the original ISSUE #1175
+		const propsClone = JSON.parse(JSON.stringify(props))
+		if (!propsClone.title) throw new Error('defineSlideMaster() object argument requires a `title` value. (https://gitbrent.github.io/PptxGenJS/docs/masters.html)')
 
 		const newLayout: SlideLayout = {
-			_margin: props.margin || DEF_SLIDE_MARGIN_IN,
-			_name: props.title,
+			_margin: propsClone.margin || DEF_SLIDE_MARGIN_IN,
+			_name: propsClone.title,
 			_presLayout: this.presLayout,
 			_rels: [],
 			_relsChart: [],
 			_relsMedia: [],
 			_slide: null,
 			_slideNum: 1000 + this.slideLayouts.length + 1,
-			_slideNumberProps: props.slideNumber || null,
+			_slideNumberProps: propsClone.slideNumber || null,
 			_slideObjects: [],
-			background: props.background || null,
-			bkgd: props.bkgd || null,
+			background: propsClone.background || null,
+			bkgd: propsClone.bkgd || null,
 		}
 
 		// STEP 1: Create the Slide Master/Layout
-		genObj.createSlideMaster(props, newLayout)
+		genObj.createSlideMaster(propsClone, newLayout)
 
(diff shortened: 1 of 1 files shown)


gitbrent:
Thanks @mikemeerschaert

PR #1177, opened 2022-10-27
bugfix: radar chart line does not respect chartColors

mathbruyen:
Option `chartColors` sets the color of markers on a radar chart but is ignored for lines which take Powerpoint's default colors. A [comment in examples](https://github.com/gitbrent/PptxGenJS/blob/a18a36a1a597e1130417741ea4d8d9363f7f5b48/demos/modules/demo_chart.mjs#L1595) states that it should be used for markers too:

> chartColors: COLORS_RYGU, // marker & line color

In case there is no dataBorder specify, fallback to chartColors.

This is very similar to bubble chart fallback (which is reindented to be be more explicit in the XML elements it creates) except that cap can be parameterized in radar.


mathbruyen:
Already fixed on master, sorry

PR #1180, opened 2022-11-07
bugfix: handle holeSize=0 for doughnut chart

mathbruyen:
Doughnut chart ignores holeSize=0 as a falsy value and takes is default of 50.

Code now checks the type of holeSize to treat 0 as a number and display doughnuts with no hole at all.

This leads to a chart very similar to a pie chart, but means that we do not have to change chart type for the 0 special case.


gitbrent (approved):

gitbrent:
Thanks @mathbruyen !

PR #1201, opened 2022-12-29
feat: Add "distributed" option to Text align options.

qsjia:
The paragraph alignment has a fifth option 'Distributed'.

http://www.addbalance.com/usersguide/justification.htm#Distributed

### DIFF
--- a/src/core-enums.ts
+++ b/src/core-enums.ts
@@ -86,6 +86,7 @@ export enum TEXT_HALIGN {
 	'center' = 'center',
 	'right' = 'right',
 	'justify' = 'justify',
+	'distributed' = 'distributed',
 }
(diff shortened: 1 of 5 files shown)


gitbrent:
Sorry, this is a Word spec not a PowerPoint spec. This code has affect on PowerPoint output.

ChatGPT:

> The OOXML (Office Open XML) presentation specification does not include a "distributed" horizontal alignment type. The horizontal alignment options in OOXML are "left," "center," "right," "justified," and "center across selection" (which centers text across merged cells).

> "Distributed" alignment is a term used in some other presentation software to describe a type of spacing where the space between characters or words is evenly distributed across a block of text. However, this type of alignment is not supported in the OOXML specification.

PR #1248, opened 2023-05-03
Fix inconsistent margins

mikemeerschaert:
Note: this PR includes breaking changes, so should probably be released for v4

### tldr;
* Changed all margins to accept inches instead of points for text 
* Changed processing order of text margin array from LRBT to TRBL so it's consistent with table margins 
* Updated demos so they supply margins in inches

### Longer description
I noticed a few inconsistencies between how table cells handle margins and how text items handle margins.
Docs for text:
image

Docs for tables:
image

for tables, the docs say you can use an array in TRBL format, and there is an undocumented feature where you can supply margins in inches, which is very helpful since the PowerPoint UI also shows margins in inches, so if you are copying a design made in PowerPoint into code you can just use the inch values. 

Text items also accept an array for margins, but it's undocumented, and processed LRBT, so even if you find the doc for table cells and deduce that it's supported for text, you're likely to be tripped up by the different order. On top of that, margins have to be in points for text, but for tables margins values < 1 are processed as inches, and >= 1 are processed as points.

Since v4 is being released soon, I recommend switching the docs to specify that margins should be supplied in inches rather than points since it's a lot more helpful, and add the margin array doc to text. This pr does that switch. I also noticed there is some code for allowing table cells to accept either points or inches, but the way the logic is structured it wouldn't work for margins over 1". That probably won't come up for table cells, but it might for text, so I don't think the logic is very sound to use for text.


lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

Issue #1168, opened 2022-09-28
[BUG|] Table Auto-Paging Error


mdbaehre:
We appreciate your feedback - to help the team understand your needs please complete the following template to ensure we have the details to help.

### Submission Guidelines

- **If you are not using the latest release, please update and see if the issue is resolved before submitting an issue**
- General questions or high-level topics should be posted in [Discussions](https://github.com/gitbrent/PptxGenJS/discussions)
- Please browse the online [Documentation](https://gitbrent.github.io/PptxGenJS/) to see if your question is already addressed there

### Issue Category

- [ ] Enhancement
- [X] Bug
- [ ] Question
- [ ] Documentation gap/issue

### Product Versions

- Please specify what version of the library you are using......: [pptxgen.bundle.js vers 3.10.0 and 3.11.0]
- Please specify what version(s) of PowerPoint you are targeting: [2019]
- Please specify what web browser you are using.................: [Chrome & Edge ]

### Desired Behavior

Create a PPTX file that does not produce a "repair" dialogue.

### Observed Behavior

The PptxGenJS code produces a PowerPoint "repair" dialogue when auto-paging a table of data that extends to 2 or more slides.  It occurs under the following conditions:
- A table row's text is split between 2 slides by the auto-paging feature AND
- The first column's text on the 2nd slide is only 1 line and does not wrap to a 2nd line in the cell AND
- The second column's text ends on the 1st slide so the cell is blank on the 2nd slide

Making the first column's text string longer so that it wraps to a second line on the second slide does not generate the powerpoint error.

Making the second column's text string longer so that it extends to the second slide also does not generate the error.

### Steps to Reproduce

A test HTML file that reproduces the error is attached below:

[autopagingErrorExample.txt](https://github.com/gitbrent/PptxGenJS/files/9669282/autopagingErrorExample.txt)

PR #1311, opened 2023-12-05
Moving the empty text fix as some were getting through before

lawtontom:
Firstly thanks for an awesome package!

We were getting intermittent errors when converting a series of Tables into powerpoint slides.  After some investigation it was down to the way the table's content was being split across slides.  There was already an effort to fill empty cells in the source code, however under certain circumstances it appears empty text entries were still in the currTableRow.

To ensure any empty entries were filled I've removed the earlier check that appeared to allow some to remain and instead loop through and fill any empty entries before finally committing the row.  I've tested this within our code base and the issue is fixed.  We'll continue to use the fix but want to contribute it back.

I believe this will also help with issues like: [https://github.com/gitbrent/PptxGenJS/issues/1168](url) and [https://github.com/gitbrent/PptxGenJS/issues/1048](url)

Happy to provide any further information as needed.  

Thanks

### DIFF
--- a/src/gen-tables.ts
+++ b/src/gen-tables.ts
@@ -469,12 +469,10 @@ export function getSlidesForTableRows (tableRows: TableCell[][] = [], tableProps
 			// 3: set array of words that comprise this line
 			const currLine: TableCell[] = srcCell._lines.shift()
 
-			// 4: create new line by adding all words from curr line (or add empty if there are no words to avoid "needs repair" issue triggered when cells have null content)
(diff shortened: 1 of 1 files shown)


lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

Issue #963, opened 2021-06-15
SVG in node.js doesn't work


Evgenus:
Hi.

I'm trying to build a presentation on the server-side (express) containing SVGs. 
Tried on versions 3.0, 3.3, 3.6 the result is the same. I've got an image inside of the presentation indicating an error. Also doesn't work in tests.  It is confusing because I see lots of issues fixed and also some remarks in comments and in the changelog.

![зображення](https://user-images.githubusercontent.com/379412/122028653-e8f40880-cdd4-11eb-8aa0-68b6efd43dc0.png)

![зображення](https://user-images.githubusercontent.com/379412/122028758-02955000-cdd5-11eb-8706-c67b3ee51891.png)

Can you please clarify if that function should work? Maybe I'm doing something wrong? Maybe missing some dependency? 

Will appreciate any reply.

tjohannto:
Maybe this helps. It works for me like this:

```
function b64EncodeUnicode(str) {
return btoa(encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, function (match, p) {
	return String.fromCharCode(+("0x" + p));
}));
}
const serializer = new XMLSerializer()
, clone = node.cloneNode(!0) //node equals the svg html node
, xmlns = "http://www.w3.org/2000/xmlns/"
, xlinkns = "http://www.w3.org/1999/xlink"
, svgns = "http://www.w3.org/2000/svg";
clone.setAttributeNS(xmlns, "xmlns", svgns);
clone.setAttributeNS(xmlns, "xmlns:xlink", xlinkns);
const dataStr = 'data:image/svg+xml;base64,' + b64EncodeUnicode(serializer.serializeToString(clone));
```
The string created I can then use in imgOptions Object.

```
const imgOptions = {};
imgOptions.x = posX;
imgOptions.y = posY;
imgOptions.w = chartWidth;
imgOptions.h = chartHeight;
imgOptions.data = dataStr ;
```

I don't know what the demo does but this creates working Presentations with v 3.4.0

gitbrent:
@Evgenus - i'm not seeing this issue, the node Image demo works fine.

Looking at the demo code, it refs 2 SVGs: a base64 one (which should always work) and a remove file:
`wikimedia_svg: { path: "https://cdn.jsdelivr.net/gh/gitbrent/pptxgenjs@master/demos/common/images/lock-green.svg" }`

```javascript
// BOTTOM-RIGHT:
[...]

PR #1315, opened 2024-01-11
Don't use IMG_BROKEN as png preview for all local / data url svg files on node

haakenlid:
Fix issue #963

The problem is that the png preview of svgs are used by LibreOffice impress, while MS PowerPoint uses the svg directly.
However, Impress shows svg files fine, and don't need the png.

The same problem shows up in web preview of the pptx file if it's shared in Slack. I guess they maybe use LibreOffice to render the preview server side.

In the demo, the svg that is downloaded from Wikipedia show up fine, because PptxGenJS simply renames the svg to png and uses it as a "preview".  However, the svg loaded from data url or from a local file will use the placeholder IMG_BROKEN png.

image

Here's the relevant part of the generated slide1.xml file for slide 1 created by the `node demo.js Image` example

Libreoffice will use the first embed "rId10", while powerpoint will display the inner one "rId11". 

```xml
<a:blip r:embed="rId10">  // <- broken png image
  <a:alphaModFix amt="50000"/>
  <a:extLst>
    <a:ext uri="{96DAC541-7B7A-43D3-8B79-37D633B846F1}">
      <asvg:svgBlip xmlns:asvg="http://schemas.microsoft.com/office/drawing/2016/SVG/main" 
                             r:embed="rId11"/> // <- correct svg image
    </a:ext>
  </a:extLst>
</a:blip>
```

This PR will make local files and data urls work the same as http url svgs. The svg will be renamed to png and used as its own preview. Which seems to have caused no problems since #528 in 2019.

However. I don't really understand the reason behind the "png preview" and the `<a:ext uri="{96DAC541-7B7A-43D3-8B79-37D633B846F1}">` tag at all. If you embed a svg using LibreOffice it will simply be included as `<a:blip r:embed="rId10">` like any other image, and it seems to display fine everywhere.

### DIFF
--- a/src/gen-media.ts
+++ b/src/gen-media.ts
@@ -107,7 +107,7 @@ export function encodeSlideMediaRels (layout: PresSlide | SlideLayout): Array<Pr
 		.forEach(rel => {
 			if (fs) {
 				// console.log('Sorry, SVG is not supported in Node (more info: https://github.com/gitbrent/PptxGenJS/issues/401)')
-				rel.data = IMG_BROKEN
+				// rel.data = IMG_BROKEN
 				imageProms.push(Promise.resolve().then(() => 'done'))
 			} else {
 				imageProms.push(createSvgPngPreview(rel))


haakenlid:
This PR should fix the bug in LibreOffice Impress, and still work fine with MS Powerpoint. 

However in DropBox's web preview of uploaded powerpoint files, the "png" will not show up correctly. I suspect this system will not accept a svg file renamed to png. 

image

But if I use LibreOffice and replace the embedded svg files with the proper svgs and save the file as .pptx, it looks good in DropBox also. 

image

Looking at the xml that LibreOffice creates, there's no .png preview file or `<a:extLst>` tag. All the xml tags used for svgs and other image formats are the same.

Which makes me wonder why PptxGenJS has this custom xml for svg files at all. Is this needed for very old versions of Microsoft Office?

```xml
<p:pic>
  <p:nvPicPr>
    <p:cNvPr id="60" name="Image 6" descr="https://cdn.jsdelivr.net/gh/gitbrent/pptxgenjs@master/demos/common/images/lock-green.svg"/>
    <p:cNvPicPr/>
    <p:nvPr/>
  </p:nvPicPr>
  <p:blipFill>
    <a:blip r:embed="rId8"/>
    <a:stretch/>
  </p:blipFill>
  [ ... ]
</p:pic>
```

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1330, opened 2024-05-07
Fix multi lvl categories in embedded .xlsx

MariusOpeepl:
## Change Summary
A key/value object was added for shared strings so that it can be used to fetch the sharedString index of category axes and removed an unneeded summand when calculating a label's column.

### Calculating the col and row position of the category labels

`getExcelColName(idx + 1 + idy)` did not need `idx`, and caused each subsequent row of labels to be shifted one more column to the right from the previous row.

### Calculating the category label indexes in sharedStrings.xml

I could not find a fix for the calculation, so I instead create a key/value object (e.g. `{ 'Category Label': 4 }`) that maps strings to their respective indexes in sharedStrings.xml, and use it to get the correct index of any given string.

### Other

There is at least one `WIP: FIXME:` note around this code; I did not remove it, as I wasn't sure if the note was referencing this issue in particular.

Lastly, I only updated the accessing of shared strings for `IS_MULTI_CAT_AXES`, as I'm not too familiar with the other chart types and I'd risk introducing more bugs. The dictionary object should get correctly populated by all chart types, though.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Motivation and Context
While the PPTX charts with multiple category axes looked fine, their embedded Excel contained malformed data, and upon opening the Excel, the related chart would also break.

- [ ] If it is a core feature, I have added new code under `/demos/modules/`
- [x] My changes generate no new eslint warnings
- [x] I have included code/tests that prove my fix is effective or that my feature works
	- The existing demos can be run to see that the fix works
- [x] I have used the "Run All Demos" feature on the [browser demo](/demos/browser/index.html) and no errors were found
	- I couldn't get the browser demos to run, so I ran the node demos instead

## Screenshots / Sample Code (if appropriate)

| Before | After |
| --- | --- |
| Microsoft_Excel_Worksheet64_broken | Microsoft_Excel_Worksheet64

### DIFF
--- a/src/gen-charts.ts
+++ b/src/gen-charts.ts
@@ -135,8 +135,8 @@ export async function createExcelWorksheet (chartObject: ISlideRelChart, zip: JS
 			)
 		}
 
-		// dictionary of shared strings mapped to indexes
-		const sharedStrings: Record<string, number> = {}
+		// dictionary of shared strings mapped to their indexes indexes
+		const sharedStringsObject: Record<string, number> = {}
 
 		// sharedStrings.xml
 		{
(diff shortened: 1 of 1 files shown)


ben-agnew:
Hey @gitbrent , I’d really appreciate it if this could be merged in. I currently need multi-level axis categories for my charts. The chart is generated correctly, but when I try to edit the data in Excel, I get a "Linked file is not available" error.

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1333, opened 2024-05-27
chore: add custGeom "custom geometry" to types for shapes

niels-bosman:
This PR adds the types needed to actually get TypeScript support when inserting custom geometry as shown in this demo: https://github.com/gitbrent/PptxGenJS/blob/master/demos/modules/demo_shape.mjs

@gitbrent Can we include this in a new release so we don’t have to monkey patch the types in our implementation?


niels-bosman:
Ping @gitbrent (!)

niels-bosman:
@gitbrent Ping :)

PR #1354, opened 2024-09-27
Fix in order to make images compatible with web workers

ghaccenta:
The code meant to detect nodejs in gen-media.ts and pptxgen.ts is mistaking web workers for nodejs and crashes. 

```ts
export function encodeSlideMediaRels (layout: PresSlide | SlideLayout): Array<Promise<string>> {
    const fs = typeof require !== 'undefined' && typeof window === 'undefined' ? require('fs') : null // NodeJS
    const https = typeof require !== 'undefined' && typeof window === 'undefined' ? require('https') : null // NodeJS

.....
```

require('fs') returns a Proxy in the context of a web worker which happens to be true-ish in the tests that are following the code snippet above. The code then tries to use fs, which obviously fails and shouldn't be used in the context of a web browser.

This pull request adds a new check to differentiate web workers from the rest of the execution :
```ts
let isWorker = false
if (typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope) {
	isWorker = true
}
```
I have also added the "webworker" lib in tsconfig.json.

Unfortunately only PNG images work, due to how SVGs are rendered (similar to node).
Function writeFile() does not work either with web workers, users should use the write() function instead.

### DIFF
--- a/src/gen-media.ts
+++ b/src/gen-media.ts
@@ -11,8 +11,13 @@ import { PresSlide, SlideLayout, ISlideRelMedia } from './core-interfaces'
  * @return {Promise} promise
  */
 export function encodeSlideMediaRels (layout: PresSlide | SlideLayout): Array<Promise<string>> {
+	let isWorker = false
+	if (typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope) {
+		isWorker = true
+	}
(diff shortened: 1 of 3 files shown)


gitbrent:
Other changes to the codebase for v4.0.0 have resolved this issue.

Also, there is a brand new WebWorker test under /demos/ now.

Thanks @ghaccenta for submitting this! Please test with the v4.0.0-rc if you can.

Issue #1299, opened 2023-10-23
PPTs failing accessibility test: Missing Table Header


bob-dlc:
The accessibility checker built into PowerPoint reports that tables are "Missing Table Header" even when I set options to {autoPage:true, autoPageRepeatHeader:true}. The top row of the table indeed acts like a header _visually_, but not _semantically_ as far as PowerPoint is concerned. As such, the PPTs fail accessibility testing and are inaccessible for users of assistive technologies such as screen readers. Are there workarounds and/or can this be fixed? Thanks.

PR #1369, opened 2025-02-05
Fix table header row for accessibility (Issue #1299)

bob-dlc:
The accessibility checker built into PowerPoint reports that all tables generated by PPTxGenJS are "Missing Table Header" even the options are set to {autoPage:true, autoPageRepeatHeader:true}. The top row of the table indeed acts like a header visually, but not semantically as far as PowerPoint is concerned. As such, the PPTs fail accessibility testing and are inaccessible for users of assistive technologies such as screen readers. 

A simple—and perhaps overly simple—solution was implemented here such that if autoPageRepeatHeader is true, the first row of the table is marked up as a header row when the XML is generated for the PowerPoint file.

Reference issue # 1299 (https://github.com/gitbrent/PptxGenJS/issues/1299)

### DIFF
--- a/src/gen-xml.ts
+++ b/src/gen-xml.ts
@@ -43,6 +43,11 @@ import {
 	valToPts,
 } from './gen-utils'
 
+// Bob Dolan 2025-01-27 | Issue #1299
+import PptxGenJS from '../types'
+import TableToSlidesProps = PptxGenJS.TableToSlidesProps
+// Bob Dolan 2025-01-27 | Issue #1299
+
 const ImageSizingXml = {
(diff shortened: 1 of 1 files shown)


pgangloff-hbs:
I reviewed the code, looks fine, I hope this will be merged soon. We need this fix for accessibility. Otherwise the generated PPT has "accessibility warnings" in MS PowerPoint. We'd rather have this PR merged into this project, to avoid creating yet another fork. Thanks!

lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1404, opened 2025-07-10
Remove unused dependencies

nwoltman:
Remove the `https` dependency because it's not used. Only the native Node.js `https` package is used. The `https` package on npm is a dummy package anyway.

Move the `@types/node` dependency to devDependencies because it's only needed for development. When this package is used in a TypeScript project, this project will have its own version of `@types/node` installed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,14 @@
 {
 	"name": "pptxgenjs",
-	"version": "4.0.1-beta.0",
+	"version": "4.0.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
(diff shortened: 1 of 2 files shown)


gitbrent:
Thanks @nwoltman !

astegmaier:
#1411 never actually made it into master, and the unnecessary `https` and `@node/types` dependencies are still in the latest public package [4.0.1](https://www.npmjs.com/package/pptxgenjs/v/4.0.1?activeTab=code). This can cause problems in monorepos that install dependencies with hoisting, because the https package effectively blocks _other_ packages from calling `import X from 'https'` - the https package is a no-op stub.

PR #1414, opened 2025-08-21
chore: remove https dependency

akinorimizushima:
Remove https that is not used in the code base.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,15 @@
 {
 	"name": "pptxgenjs",
-	"version": "4.0.1-beta.0",
+	"version": "4.0.1",
 	"lockfileVersion": 3,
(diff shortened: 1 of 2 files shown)


gitbrent:
Thanks @akinorimizushima 

This has already been done. Closing as duplicate of Pull #1404

astegmaier:
#1404 never actually made it into master, and the unnecessary `https` and `@node/types` dependencies are still in the latest public package [4.0.1](https://www.npmjs.com/package/pptxgenjs/v/4.0.1?activeTab=code). This can cause problems in monorepos that install dependencies with hoisting, because the https package effectively blocks _other_ packages from calling `import X from 'https'` - the https package is a no-op stub.

PR #1427, opened 2025-11-06
Add missing XML encoding to company attribute

hhulkko:
Fixes error "PowerPoint found a problem with content in [path to generated file]" when opening the generated file if there are xml character entities in the "company" attribute.


lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

Issue #1436, opened 2026-01-20
[BUG] Scatter and bubble chart inherit the same value formatting


Ben-vD:
### Issue Category

Bug

### Product Versions

- Please specify what version of the library you are using......:[4.0.1.]
- Please specify what version(s) of PowerPoint you are targeting: [365]
- Please specify what web browser you are using.................: [Chrome, Firefox]

### Desired Behavior

For the scatter and bubble chart, I want to be able to format the values on both the x and y axis independently using the catLabelFormatCode and valAxisLabelFormatCode parameters respectively.

### Observed Behavior

Currently, both the x and y axis for the scatter and bubble chart inherit the format provided to the valAxisLabelFormatCode

### Steps to Reproduce

Create a scatter or bubble chart using the PptxgenJS framework. Provide a valAxisLabelFormatCode, which formats both the x and y axis. Thus it is not possible to format the x and y axis differently.

PR #1438, opened 2026-01-20
[BUG FIX] Scatter and Bubble chart formatting fix

Ben-vD:
I have changed the categorical axis formatting for the scatter and bubble charts to use the catLabelFormatCode parameter. The original implementation used valAxisLabelFormatCode, which is also applied to the value axis. Since both parameters use standard formatting encodings, the categorical axis (x-axis) for scatter and bubble charts can instead use the catLabelFormatCode parameter. This allows for separate formatting of the x- and y-axes for scatter and bubble charts. This fixes issue #1436.

I have only included the src/ file to prevent any conflicts. Let me know if I need to run the pipeline locally and update the other files.


lofcz:

Change dependency:
```
npm uninstall pptxgenjs
npm i pptxgenjs-plus
```

And update your imports:
```diff
- import pptxgen from "pptxgenjs"
+ import pptxgen from "pptxgenjs-plus"
```

PR #1455, opened 2026-04-20
fix: inner shadow property breaks PPTX file

### LINKED ISSUE #1293: Inner shadow property breaks pptx file
I believe I have found an unusual bug. 

The shadow property breaks the pptx file if the type is set to 'inner', whereas it would generate the file soundly if it were set to 'outer'. 

This is not to say that the inner shadow is not working for all components, as this has been noticed with rounded rectangle and rectangle shapes.

you can test it yourself with this example:
```
let pptx = new PptxGenJS();
let slide = pptx.addSlide();

var sh_type = 'outer' // try changing it to inner
var shadowOpts = { type: sh_type , color:'000000', blur:1, offset:0, angle:0, opacity:0.3 };


slide.addShape(pptx.ShapeType.roundRect, {
  x: 0.0,  y: 0.25,  w: '100%',  h: 1.5,
  align: 'center',  fontSize: 24,
  color: '0088CC',  fill: 'F1F1F1',
  rectRadius: 45,
  shadow: shadowOpts
});

// Save the PowerPoint file
pptx.writeFile("example.pptx")
```

dxsun97:
Fix #1293 

Root cause: In src/gen-xml.ts:531, the shape shadow closing XML tag is hardcoded as </a:outerShdw> instead of using the dynamic shadow type. When 'inner' is used, the XML opens with <a:innerShdw> but closes with </a:outerShdw>, creating malformed XML that PowerPoint cannot parse.


ShubhamVsCode:
I think the library is not maintained properly, the last commit in the main is from 10 months ago, i also got this same issue

Is there any actively maintained pptx gen library?

dxsun97:
> I think the library is not maintained properly, the last commit in the main is from 10 months ago, i also got this same issue
> 
> Is there any actively maintained pptx gen library?

I'm trying to create a new lib for it👨🏻‍💻

PR #1528, opened 2026-08-20
ci: dispatch-driven release with automatic version bump

MBuelowius:
Releasing no longer requires editing version strings by hand.

## What changed

**`release.yml` gains a `workflow_dispatch` path.** Run the Release workflow on `master`, pick `patch` / `minor` / `major`, and after the usual `release` environment approval one job does everything: bump + tag, rebuild `dist/`, verify the bundles, `npm publish`, create the GitHub Release with generated notes, upload the bundles.

The existing `release: published` trigger is untouched, so cutting a release by hand (including betas) still works exactly as before.

**Why one job instead of a separate bump workflow.** A tag pushed with `GITHUB_TOKEN` cannot trigger another workflow — a standalone bump workflow would push the tag into silence. Doing the whole release in the approved job avoids that without needing a PAT. For the same reason, the final `gh release create` does not re-trigger the published path, so there is no double publish.

Ordering matters and is deliberate: the bump runs before `npm ci` so the new version bakes into the dist banner, and the tag push happens before `npm publish` so a rejected push aborts before anything reaches npm.

**`scripts/sync-version.mjs`** is wired to the `version` npm lifecycle hook, so `npm version` also rewrites `const VERSION` in `src/pptxgen.ts` and the `types/index.d.ts` header, and folds them into the same commit. It throws if either version line stops matching rather than silently shipping a stale constant. Verified end to end in a scratch repo: `npm version patch` produces one commit containing all three files plus the tag.

**`RELEASING.md`** documents the new path and marks the manual steps it replaces.

## Contingency

If `master` is branch-protected, the bot's push will be rejected. It fails early, before npm publish, so the abort is clean — the fix is a PAT or GitHub App token with bypass. Not pre-built here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MBuelowius:
Opened against the wrong repository by mistake — this belongs on the NeomaVerwaltung fork. Sorry for the noise.

